PR #3, opened 2015-11-30
chore(Makefile): switch to alpine-based dev environment

krancour:
The alpine-based dev env is experimental, but so far seems to work nicely with the router.  Let's roll with it for now.

Ping @helgi @technosophos

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,7 +8,7 @@ BUILD_TAG := git-${GIT_SHA}
 
 # The following variables describe the containerized development environment
 # and other build options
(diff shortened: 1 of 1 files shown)


helgi:
Off hand any limitations beyond lack of `CGO`?

krancour:
Allegedly the libc library is non-standard, but since we're not actually cross-compiling, I'm anticipating that doesn't end up mattering much.  @technosophos can speak to this better than I.  So far, however, this _is_ working for the router.  Want to let this marinate for a while.  If we go back on it eventually, so be it.  This is alpha-- now is the time for these experiments.  :)

PR #5, opened 2015-12-01
feat(router): send nginx logs to stdout and stderr

krancour:
The PR also includes some fixes that were necessary pre-requisites to this.

### DIFF
--- a/nginx/commands.go
+++ b/nginx/commands.go
@@ -2,12 +2,20 @@ package nginx
 
 import (
 	"log"
+	"os"
 	"os/exec"
+)
+
+const (
+	nginxBinary = "/opt/nginx/sbin/nginx"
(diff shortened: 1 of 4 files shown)


bacongobbler:
did this resolve the logging error you were seeing last night? :)

krancour:
Yes.  It seems there was a confluence of factors.  Each commit is targeted at one of the factors.  Previous to this PR, Nginx worker processes were running as `nobody` (which cannot have been good in terms of permissions), were not using `daemon off` (which has been documented as required when running Nginx in a container), and were launching Nginx using an intermediate shell process (this was ugliness inherited from the k8s/contrib nginx-alpha ingress controller).  Fixing these things got the logs flowing.

PR #6, opened 2015-12-01
fix(Makefile): check for dev registry before build

krancour:


bacongobbler:
just realized that this is necessary so the image name is fully formed. +1

PR #9, opened 2015-12-01
chore(manifests): remove superfluous app label on rc; add heritage

krancour:


bacongobbler:
I know that we use the app label still, but I agree that it is superflous since the `name` metadata is the exact same thing, so LGTM.

technosophos:
Please put the `app` label back on. That label is used by some k8s admin tools, and we're trying to make sure we are compatible.

http://kubernetes.io/v1.1/docs/user-guide/walkthrough/k8s201.html

technosophos:
@krancour pointed out offline that RCs do not have the app label. Only pods and pod templates.

krancour:
Thanks @technosophos

PR #10, opened 2015-12-01
chore(manifests): replace blank image attribute with comment

krancour:


iancoffey:
I think every component handles this in its own way - it would be nice to arrive at one method eventually.

krancour:
@iancoffey I took this approach from deis/workflow.  The only difference is that I might have renamed the target that does it to make it more "familiar" by resembling v1.x target names.  Eventually, people may decide they don't like this, but I've been trying to make router conform to that old familiar developer workflow as much as I can-- even if the underlying implementation of each target has changed a bunch.

iancoffey:
:+1:

PR #15, opened 2015-12-02
feat(ci): integrate with travis

krancour:
Requesting reviews from @helgi and @mboersma

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,18 @@
+branches:
+  only:
+    - master
+cache:
+  directories:
+    - vendor
+sudo: required
+services:
+  - docker
+install:
+  - make bootstrap
+script:
+  - make test
+deploy:
+  provider: script
+  script: _scripts/deploy.sh
+  on:
+    branch: master
--- a/Makefile
+++ b/Makefile
@@ -1,10 +1,8 @@
 include includes.mk
 
 SHORT_NAME := router
-VERSION := 2.0.0-$(shell date "+%Y%m%d%H%M%S")
-
 GIT_SHA := $(shell git rev-parse --short HEAD)
-BUILD_TAG := git-${GIT_SHA}
+VERSION ?= git-${GIT_SHA}
 
 REPO_PATH := github.com/deis/${SHORT_NAME}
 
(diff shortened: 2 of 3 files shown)


krancour:
Well... this appears to work, but I just realized I am missing the special sauce to clearly tag these as v2-alpha builds.  Fixing that now.

krancour:
Ok.  Should be better now.

krancour:
But probably needs re-review.

.travis.yml
```diff
@@ -0,0 +1,18 @@
+branches:
+  only:
+    - master
+cache:
+  directories:
+    - vendor
```
mboersma: Is using a local directory here sufficient? In the other .travis.yml files we have made the path explicit: ` $GOPATH/src/github.com/deis/router/vendor`. I've just copied that and assumed it was necessary since I'm unclear what `PWD` is on Travis.

Makefile
```diff
@@ -14,7 +14,7 @@ DEV_ENV_IMAGE := quay.io/deis/go-dev:0.2.0
 DEV_ENV_WORK_DIR := /go/src/${REPO_PATH}
 DEV_ENV_CMD := docker run --rm -v ${PWD}:${DEV_ENV_WORK_DIR} -w ${DEV_ENV_WORK_DIR} ${DEV_ENV_IMAGE}
 DEV_ENV_CMD_INT := docker run -it --rm -v ${PWD}:${DEV_ENV_WORK_DIR} -w ${DEV_ENV_WORK_DIR} ${DEV_ENV_IMAGE}
-LDFLAGS := "-s -X main.version=${VERSION}"
+LDFLAGS := "-s -X main.version=${BUILD_TAG}"
```
mboersma: Actually I think `VERSION` is preferred here, at least if we allow the other components a vote. deis/workflow is the only one that uses `BUILD_TAG` instead, and it should probably be changed.

.travis.yml
krancour: According to examples in Travis' own documentation, this is sufficient.  Also keep in mind that I'm trying to not _be_ a "go project" since I'm using a fully containerized dev/build/test environment.  Outside the container, `$GOPATH` isn't a thing for me.

Makefile
krancour: I'll change this.

.travis.yml
```diff
@@ -0,0 +1,18 @@
...
+sudo: required
+services:
+  - docker
+install:
+  - make bootstrap
+script:
+  - make test
+deploy:
+  provider: script
+  script: make ci-deploy
```
mboersma: Should this be an external script like _scripts/deploy.sh in the other components? The Makefile target could just call that...

krancour: Personal preference-- felt things were a little simpler by keeping all the deploy logic in one place.  (i.e. Deploy to local cluster when developing and CI deploy are both described by make targets.)  If there's a very compelling reason not to do this, I would be happy to reconsider that.

mboersma: The arguments for a separate deploy.sh would be:
- keeps the Makefile simpler
- that's what we're doing in deis/etcd, deis/workflow, deis/builder, deis/slugbuilder, and so on--they all have an identical _scripts/deploy.sh
- this is the quick-and-dirty "get an image up on Docker Hub" version, but very soon the logic will need to expand to include versioning and tags, and then we'll be glad it's consistently implemented
- hopefully we can stop the copy-paste and include deploy.sh from a central location at some point
- @sgoings thought this was the right initial approach

krancour: Ok.  I'm sufficiently convinced.

krancour:
@mboersma your last bit of feedback has been addressed.  The only differences between my script and deis/etcd's (where I copied it from) is that my make target it calls is just `push` instead of `docker-push`.  If all of us want to get together and hash out our opinions on how to name targets consistently across components, that's probably an entirely separate conversation that I wouldn't want to hold up this PR... but it probably needs to happen at some point.

PR #18, opened 2015-12-03
feat(router): support stream for builder

krancour:
Attn: @arschles 

While adding support for the builder, I stumbled across the edge case of missing annotations.  In the case of builder, I can roll with defaults.  Awareness of this edge case and my technique for handling it did lead to a more general refactor of how this chunk of code builds the model, that and a liberal sprinkling of new comments to explain what the hell I'm doing.  Please forgive that slight scope creep.

~~If you want to use this with builder (as installed by helm), there's only one thing you actually need to do-- you need to move the builder's service into the "deis" namespace.  Moving all the deis components into the "deis" namespace has been a previous topic of discussion and has been something we'll "get around to at some point."  I guess it's time?  (It sort of matters to the router, because the router goes _looking_ for deis-builder and it really does need to know what namespace to look in.)  We could discuss this at greater length tomorrow.  cc: @technosophos @mboersma~~ This was addressed by https://github.com/deis/charts/pull/12

~~Apart from that, the remainder of~~ the configuration is optional.  You can, but don't have to, do the following in the manifest for the deis-builder:

```
apiVersion: v1
kind: Service
metadata:
  annotations:
    routerConfig: '{"connectTimeout":10000,"tcpTimeout":1200000}'
...
```

The values shown are the defaults you get if you don't add a `routerConfig` annotation.

### DIFF
--- a/model/model.go
+++ b/model/model.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 
 	"k8s.io/kubernetes/pkg/api"
+	"k8s.io/kubernetes/pkg/api/errors"
 	client "k8s.io/kubernetes/pkg/client/unversioned"
 	"k8s.io/kubernetes/pkg/labels"
 )
@@ -12,6 +13,7 @@ import (
 type RouterConfig struct {
 	UseProxyProtocol bool `json:"useProxyProtocol"`
 	AppConfigs       []*AppConfig
+	BuilderConfig    *BuilderConfig
 }
 
 func newRouterConfig() *RouterConfig {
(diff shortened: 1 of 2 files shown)


krancour:
See https://github.com/deis/charts/pull/12

model/model.go
```diff
@@ -31,18 +33,43 @@ func newAppConfig() *AppConfig {
...
+	}
+}
+
 // Build creates a RouterConfig configuration object by querying the k8s API for
 // relevant metadata concerning itself and all routable services.
 func Build(kubeClient *client.Client) (*RouterConfig, error) {
-	rc, err := getRC(kubeClient)
+	// Get all relevant information from k8s:
+	//   deis-router rc
+	//   All services with label "routable=true"
```
arschles: this label is new to me. I'm assuming we'll need to add it to all applicable services (`deis-builder`, `deis-workflow`, etc...)? also, does that include user apps?

krancour: I'm building explicit support for builder and workflow.  It's just to find user apps that need traffic routed to them.

arschles: got it. thanks

arschles:
is this ready for testing in a cluster?

krancour:
It's not feature complete.  It will, however, do everything it needs to with respect to the builder.

arschles:
ok. I'll be happy to test it in the near future

PR #19, opened 2015-12-03
feat(Makefile): Dev Environment container inception

hermanjunge:
Enable users to bootstrap the development environment from a container.
Use case: A developer has their tools inside a container and mounted
this directory from the host in it.


hermanjunge:
OK. This is not a little, but completely crazy. Fact is, I'm experimenting with a completely remote [containerised development environment](https://github.com/hermanjunge/dockerfiles/blob/master/ubuntu-dev-docker/Dockerfile) over a coreOS machine in a droplet.

Has its benefits, the whole `glide` operation takes 2'20'' to complete there, as opposed to an _eternity_ using my `4Mbps` home connection :sleeping:. You just have to mount your docker socket and play from it. The problem starts when your _inside container_ router directory (ex: `/root/dev/router`) differs from the _outside container_ router directory (ex:`/home/core/dev/router`), then, docker would not be able to mount your directory, stopping you from the joy of getting all glide dependencies fast.

Using this variable I added, `DEV_ENV_HOST_DIR`, fixes that and enables this little inception.

krancour:
I've got to think about this a little bit.  This seems like an extreme outlying edge case.

arschles:
@hermanjunge I don't quite get it. Are you spinning up a container in a pod and then ssh-ing to that container or something?

krancour:
@arschles he has a container running on a remote host that has `make`, `docker` (in docker), and `kubectl`.  So while I have containerized go and such, he's containerized everything.  (Of course, your local machine still needs a docker client.)  While this is neat, I think the "inception" nature of it complicates things a good bit.  I think I'd want t see some sort of evidence that many others want to work in this way before adding any explicit support for it.

sgoings:
I think this is a neat idea. Though difficult (and messy) to do now... wouldn't that be the dream? From any machine spin up your dev environment per project (think rbenv but where your build environment is remote on AWS, GCE, etc.).

Right now it makes more sense to me to just spin up a virtual machine in a cloud provider to do this (and avoid the docker-in-docker), but that requires quite the setup and backing code to do so.

krancour:
To be clear, I don't want to stop anyone from doing this, but I'm leaning towards not explicitly supporting it.

sgoings:
Makes sense, @krancour. To me it feels very much like what [Wercker](http://wercker.com/) is headed toward, but with the idea of a whole development workflow baked in.

hermanjunge:
I agree with @krancour that this is a _long_ shot, and with @sgoings that is way easier to just do everything in a remote machine well configured to do that instead of a container. Actually I started with an ubuntu virtual machine, and then with a container over coreOS inside a droplet.

This PR has more an intention to share the finding in the weird case something else needed a heads up on this particular. So is perfectly reasonable to close this PR, and maybe just add the note in some sort of "Special Use Case" document, with a disclaimer of no support, of such.

arschles:
@krancour @hermanjunge right, but you still have to write code, right? do you write it inside the container?

hermanjunge:
@arschles No. Besides the `Dockerfile`, and the modification in this PR, no other code is needed to make this proposal work. You can edit the code of `deis/router` outside or insider the container, it does not make any difference, since you are mounting it as a volume. (If I understood your question well).

arschles:
yes, I think you understood right. so you're using the VM's filesystem to store your code and mounting it? If so, fwiw I think this is a fine idea. I've used a similar dev workflow (packaging up all tools in an VM image, developed on big codebase on NAS, submitted patches to a build server) and it certainly had advantages

hermanjunge:
In my particular use case, I don't have the bandwidth I need to play fast with repositories, apt packages and docker images, so what I do is create a Virtual Machine in the cloud, load it with the adequate environment for the task, perform the development, push back and destroy the VM.

Your mileage may vary, but from where I'm standing, this is a huge time saver.

arschles:
I'm :+1: on the time saver point also.

@krancour @sgoings perhaps this only changes 2 vars in the makefile. perhaps we could move those changes to under a build target so we can make this work w/o altering the current workflow?

technosophos:
As we're getting down to the Alpha milestone freeze, would someone mind weighing in on whether this needs to go into Alpha?

krancour:
It does not.

hermanjunge:
I'd close the issue. As @krancour stated, it represents an extreme edge case. Anyways, It is good enough as a reference for anyone working on something similar.

PR #20, opened 2015-12-04
chore(manifest): improve formatting of structured data in routerConfig annotation

krancour:


krancour:
Just to address the obvious question of why I'd embed JSON inside YAML at all...

Kubernetes parses `metadata.annotations` as `map[string]string`.  If the values were to contain any structure of their own, it's not parsed automatically-- meaning I've got to do it myself.  So having a single annotation `routerConfig` that is a `string` containing JSON, the router's Go code can easily parse all the configuration into a model object.  Go's got first class support for parsing JSON, so this actually works out pretty well.  New router settings can be accommodated with absolutely minimal code changes as router development continues.

helgi:
Would be nicer if `annotations` could have a structure, this looks just a bit out of place. hrmm. Is there any upstream talks about doing something like that or are they pretty solid on keeping `annotations` as string?

krancour:
Per their docs, annotations can be structured data, but my experimentation shows what they _mean_ by that is annotations are a `map[string]string` (which _is_ a structure, eh?), but any structure beyond that, you're responsible for yourself.

If this could or should change upstream and we were to then leverage it, it would wreck any chance of compatibility with k8s 1.1.

I know this looks odd, but let's not get hung up on it.  My previous comment was to explain the oddity, but this OR does not _introduce_ said oddity-- it merely formats it differently.

PR #21, opened 2015-12-04
chore(manifest): add more port mappings

krancour:


manifests/deis-router-rc.yaml
```diff
@@ -23,3 +23,7 @@ spec:
         ports:
         - containerPort: 80
           hostPort: 80
+        - containerPort: 443
+          hostPort: 443
+        - containerPort: 2222
```
helgi: Do we use `2222` for anything beside builder? Wondering how (if?) manifests are merged together on the fly and if the builder could own that port opening instead, as in instruct router to open it up. Probably a bit too complicated for now tho.

krancour: Builder would have 2222 open within its container, and a service resource provides a VIP:2222 that maps to that.  That's actually how all the components work.  Router is an exception to that.  Since router is meant to be the single point of ingress for the Deis platform, router needs port mappings firectly from the host into its container.  This makes router unique among the components.... it will directly answer traffic received by the host on these ports.  This would allow, for instance, a load balancer outside the cluster to front (potentially) _all_ the nodes, but through a health check, determine which ones can actualky receive traffic.  This does differ a bit from how k8s _wants_ to do things.  It would rather, for instance that we expose a service of `type: LoadBalancer` for the router, which would cause k8s to use the applicable "cloud provider" (which is really an _adapter_ to your underlying infrastructure provider) to create the load balancer for you.  Ultimately, we may want to do that still, but on Vagrant or bare metal, for instance, the bits that manage a load balancer external to the cluster are a no op... which forces our hand.  We have to employ the host-> container port mapping strategy seen here to provide a point of ingress that's guaranteed to work universally.

krancour: Could builder do that host:2222 -> container:2222 port mapping instead?  Yep... but then there'd be two points of ingress into Deis (2222 and 80/443), which would necessitate _two_ load balancers external to the cluster and some fancier DNS records that direct builder traffic to the second LB.  Bottom line here is the strategy I'm employing is actually identical to the one used in 1.x.

helgi: What I was aiming for here is can `builder` append to the `host -> port` mapping from its manifest / chart so that the `router` doesn't have to? Or are there limitations on how much `rc`s can affect one another like that?

This isn't a `builder` specific question per se and more if X arbitrary chart can add to the mapping to expose different ports on the router for various random needs.

helgi:
I wouldn't want 2 load balancers - just if we can somehow affect the configuration on *one* LB via multiple Charts :)

manifests/deis-router-rc.yaml
helgi: I wouldn't want 2 load balancers - just if we can somehow affect the configuration on *one* LB via multiple Charts :)

krancour: Forget charts.... charts are just k8s manifests underneath.  So the question is one of whether an RC can modify the `template.spec` of _another_ RC.  Nope.  And I'm glad it can't.  That would be very confusing, not to mention, it could also potentially be exploited.

helgi: charts/manifests - it's all mixing together :D

I understand it can be exploited but I was hoping k8s had some niceness in that area. Having a system that could say "I'd like 2222 to be open to the world, who is in charge of Ingress, please make it happen" and the singular LB would be able to respond to that instead of needing a new LB (i'm ignoring how each component would auth against one another in that case)

Idle thoughts since I'm not a fan of having to specify a builder port on a router RC

krancour: > who is in charge of Ingress, please make it happen

This concept exists in k8s.  It's actually _called_ ingress, but it's beta, and available implementations of "ingress controllers" are alpha.  (This router is inspired by the nginx-alpha ingress controller.)

And this would be a big part of why ingress isn't stable yet-- it doesn't at all address how ingress for anything other than 80/443 is handled... yet.  It is meant to eventually.

PR #23, opened 2015-12-08
feat(router): make platform domain configurable

krancour:
Before this PR, routable services could be annotated with a list of domains, but that meant if someone wanted to change their "platform domain," every routable service would require its annotations to be updated-- no bueno.

This PR allows the platform domain (e.g. example.com) to be configured at the _router_ level and any "domains" defined on a routable application that do not contain any periods will be assumed to be subdomains of the platform domain when the router builds its configuration model.

### DIFF
--- a/manifests/deis-router-rc.yaml
+++ b/manifests/deis-router-rc.yaml
@@ -8,6 +8,7 @@ metadata:
   annotations:
     routerConfig: |
       {
+        "domain": "example.com",
         "useProxyProtocol": false
       }
 spec:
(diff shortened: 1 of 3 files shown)


mboersma:
This LGTM, but I wish there were a `model_test.go` with friends. :green_heart:

arschles:
+1 to @mboersma. added #25 but not blocking this PR - gave LGTM

krancour:
@mboersma there will be.  I'm in a race for feature parity with 1.x to avoid blocking any other end-to-end testing.  I still intend to backfill tests pre-alpha.

PR #24, opened 2015-12-08
chore(router): prefix annotations with deis.com/

krancour:
Suggested by @technosophos


krancour:
Merge conflicts fixed.

helgi:
We want `.com` instead of `.io` ?

bacongobbler:
I'm +1 on using .io since that's the open-sourcey side of Deis where .com is the commercial support side. Makes more sense to differentiate between the two (then if we have commercial apps in the future we could use deis.com annotations)

krancour:
I defer to @technosophos on this.  It was his suggestion.

helgi:
cc @slack

krancour:
fwiw, I'd throw my vote to `deis.io` too.

slack:
I don't mind `.io`

krancour:
Switched to `.io`

krancour:
Going to treat the LGTMs as if they still stand and proceed with merging after CI passes, unless anyone speaks up between now and then.

PR #27, opened 2015-12-08
doc(router): document config options

krancour:
This is a rough cut of instructions for how to configure the router via annotations.  Right now this is mainly for @helgi's benefit, especially WRT to the router configuration options he can make workflow add/manage on each application's k8s services.

More options will be added to this table as they are implemented.

### DIFF
--- a/README.md
+++ b/README.md
@@ -127,6 +127,124 @@ When generating configuration, the program parses structured data (JSON) found i
 
 Similarly, the router watches its _own_ `routerConfig` annotations to dynamically construct global Nginx configuration.
 
+## Configuration Guide
+
+### Environment variables
+
+Router configuration is driven almost entirely by annotations on the router's replication controller and the services of all routable applications-- those labeled with `routable=true`.
+
+One exception to this, however, is that in order for the router to discover its own annotations, the router must be configured via environment variable with some awareness of its own namespace.  (It cannot query the API for information about itself without knowing this.)
+
+The `POD_NAMESPACE` environment variable is required by the router and it should be configured to match the Kubernetes namespace that the router is deployed into.  If no value is provided, the router will assume a value of `default`.
+
+For example, consider the following Kubernetes manifest.  Given a manifest containing the following metadata:
+
+```
+apiVersion: v1
+kind: ReplicationController
+metadata:
+  name: deis-router
+  namespace: deis
+# ...
+```
+
+The corresponding template must inject a `POD_NAMESPACE=deis` environment variable into router containers.  The most elegant way to achieve this is by means of the Kubernetes "downward API," as in this snippet from the same manifest:
+
+```
+# ...
+spec:
+  # ...
+  template:
+    # ...
+    spec:
+      containers:
+      - name: deis-router
+        # ...
+        env:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -127,6 +127,122 @@ When generating configuration, the program parses structured data (JSON) found i
...
+  # ...
+  template:
+    # ...
+    spec:
+      containers:
+      - name: deis-router
+        # ...
+        env:
+        - name: NAMESPACE
+          value: deis
```
bacongobbler: we can use the downward API for this: https://github.com/deis/charts/blob/master/deis/manifests/deis-etcd-discovery-rc.yaml#L34-L37

README.md
```diff
@@ -127,6 +127,122 @@ When generating configuration, the program parses structured data (JSON) found i
 
 Similarly, the router watches its _own_ `routerConfig` annotations to dynamically construct global Nginx configuration.
 
+## Configuration Guide
+
+### Environment variables
+
+Router configuration is driven almost entirely by annotations on the router's replication controller and the services of all routable applications-- those labeled with `routable=true`.
```
bacongobbler: should we change this annotation to `deis.io/routable`, or is it fine to leave it without a domain?

README.md
```diff
@@ -127,6 +127,122 @@ When generating configuration, the program parses structured data (JSON) found i
...
+
+| Component            | Name             | Type       | Default Value | Description |
+|----------------------|------------------|------------|---------------|-------------|
+| deis-router          | domain           | `string`   | N/A           | This defines the router's default domain.  Any domains added to a routable application _not_ containing the `.` character will be assumed to be subdomains of this default domain.  Thus, for example, a default domain of `example.com` coupled with a routable app counting `foo` among its domains will result in router configuration that routes traffic for `foo.example.com` to that application. |
+| deis-router          | useProxyProtocol | `boolean`  | `false`       | PROXY is a simple protocol supported by nginx, HAProxy, Amazon ELB, and others.  It provides a method to obtain information about a request's originating IP address from an external (to Kubernetes) load balancer in front of the router.  Enabling this option allows the router to select the originating IP from the HTTP `X-Forwarded-For` header. |
```
bacongobbler: +1 for disabling by default

README.md
krancour: Yup.  I actually have PR'ed that in this repo's manifests and it's already fixed in the charts, but overlooked it in the docs.  Nice catch.

README.md
krancour: That is in reference to a label; not an annotation.  If we want to start prefixing label keys, that's a broader discussion because we're not currently prefixing the `app` or `heritage` labels, for instance, although I suppose an argument can be made that those two labels are special.

bacongobbler: gotcha.

krancour:
@bacongobbler I see you already LGTM'ed (thanks), but just FYI... the bit of doc that could have demonstrated use of the downward API to set the `POD_NAMESPACE` environment variable now does.

bacongobbler:
+1

arschles:
fwiw LGTM also. @carmstrong beat me to the label :)

PR #28, opened 2015-12-09
chore(router): use downward api to set POD_NAMESPACE env var

krancour:
This is insipired by deis/charts#23

For consistency, I also renamed the env variable to `POD_NAMESPACE`

### DIFF
--- a/manifests/deis-router-rc.yaml
+++ b/manifests/deis-router-rc.yaml
@@ -25,8 +25,10 @@ spec:
         image: ## Image is dynamically set by `make set-image` as part of dev deployment ##
         imagePullPolicy: Always
         env:
-        - name: NAMESPACE
-          value: deis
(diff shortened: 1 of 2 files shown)


manifests/deis-router-rc.yaml
```diff
@@ -25,8 +25,10 @@ spec:
         image: ## Image is dynamically set by `make set-image` as part of dev deployment ##
         imagePullPolicy: Always
         env:
-        - name: NAMESPACE
-          value: deis
+        - name: POD_NAMESPACE
+          valueFrom:
+            fieldRef:
+              fieldPath: metadata.namespace
```
technosophos: :boom: Downward API for the win!

krancour: I am loving this feature.

PR #30, opened 2015-12-09
feat(router): configure "basic settings"

krancour:
This PR continues the march toward feature parity.  ~~WIP, because it will need a rebase and doc updates after #27 is merged.~~

Note the "basic settings" are inherited directly from the v1.x router.

### DIFF
--- a/nginx/config.go
+++ b/nginx/config.go
@@ -16,6 +16,13 @@ events {
 }
 
 http {
+	# basic settings
+	vhost_traffic_status_zone;
+
+	sendfile on;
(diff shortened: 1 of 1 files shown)


nginx/config.go
```diff
@@ -16,6 +16,13 @@ events {
 }
 
 http {
+	# basic settings
+	vhost_traffic_status_zone;
```
mboersma: Some vts reference here, for my own edification: https://github.com/vozlt/nginx-module-vts#vhost_traffic_status_zone

PR #34, opened 2015-12-09
feat(router): make hash sizes configurable

krancour:
Another incremental step toward feature parity.


krancour:
Worth noting: this PR adds configuration options that are nginx specific.  After feature parity has been achieved, I intend to circle back and add support for configuration "extensions" that might be implementation-specific.  See #36.

PR #35, opened 2015-12-09
feat(router): add gzip support

krancour:
Add gzip support.  ~~WIP because this requires rebase and additions to the docs after #27 is merged.~~


README.md
```diff
@@ -185,6 +185,14 @@ Note that although the annotation containing router configuration for each of th
 | deis-router          | workerProcesses  | `string`   | `auto` (number of CPU cores) | Number of worker processes to start. |
 | deis-router          | workerConnections | `integer` | `768`         | Maximum number of simultaneous connections that can be opened by a worker process. |
 | deis-router          | defaultTimeout   | `integer`  | 1300          | Default timeout value in seconds.  Should be greater than the front-facing load balancer's timeout value. |
+| deis-router          | useGzip          | `boolean`  | `true`        | Whether to use gzip compression. |
```
helgi: is it a limitation in `confd` that wouldn't allow you to test if `gzipConfig` is empty or not and thus apply whichever of the config elements?

krancour: Not using confd at all currently.  But the reason for the existence of `useGzip` is that I initialize a nested `GzipConfig` with all the defaults gzip settings.  Since I initialize that with some default state, the nested `GzipConfig` is _never_ `nil`.  Hence, a separate switch to turn the whole thing on or off.

It might be possible to allow an operator to `nil` out the whole nested `GzipConfig`.  For my own edification, I am going to try that, but I'm not sure it's a more intuitive means of disabling gzip than by setting `"useGzip": false`.

helgi: oh I see that now - You make a valid point about the default values there and as such `useGizp` makes sense if I wouldn't want it at all, just from the outside it feels like an extra argument to deal with but its on by default so no big deal. Maybe we can explore different approaches when you are happy with alpha :)

PR is ready either way

krancour: Ok... this can work... this in the annotations:

```
metadata:
  annotations:
    deis.io/routerConfig: |
      {
        "domain": "example.com",
        "useProxyProtocol": false,
        "gzipConfig": null
      }
```

And this simple change in the template:

```
{{ if $routerConfig.GzipConfig }}{{ $gzipConfig := $routerConfig.GzipConfig }}gzip on;
```

I guess that with the proper docs, this is a reasonable means of wiping out the default configuration to completely disable gzip.

I'm going to make that change.

krancour:
Closing in favor of #38

PR #38, opened 2015-12-09
feat(router): add gzip support (option 2)

krancour:
Alternative implementation of #35.  Ping @helgi: this drops the extra `useGzip` config option in favor of operators nulling out the `gzipConfig` if they want to disable it entirely.  lmk what you think.


helgi:
Yeah this is what I had in mind. Looking good :+1:

krancour:
@helgi thanks for the suggestion.  I was hesitant at first, but in the end, I like this better.

PR #43, opened 2015-12-10
chore(router): inherit more app proxy settings from v1.x

krancour:
Fill out the remaining proxy settings for an app, minus anything having to do with SSL, enforcing HTTPS, etc.

### DIFF
--- a/nginx/config.go
+++ b/nginx/config.go
@@ -52,6 +52,11 @@ http {
 	access_log /opt/nginx/logs/access.log upstreaminfo;
 	error_log  /opt/nginx/logs/error.log error;
 
+	map $http_upgrade $connection_upgrade {
+		default upgrade;
+		'' close;
+	}
+
 	# Default server handles requests for unmapped hostnames
 	server {
 		listen 80{{ if $routerConfig.UseProxyProtocol }} proxy_protocol{{ end }};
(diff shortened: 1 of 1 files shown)


nginx/config.go
```diff
@@ -65,8 +70,16 @@ http {
 		listen 80{{ if $routerConfig.UseProxyProtocol }} proxy_protocol{{ end }};
 		server_name {{$domain}};
 		{{ if $appConfig.Available }}location / {
+			proxy_buffering off;
+			proxy_set_header Host $host;
+			proxy_set_header X-Forwarded-For {{ if $routerConfig.UseProxyProtocol }}$proxy_protocol_addr{{ else }}$proxy_add_x_forwarded_for{{ end }};
+			proxy_redirect off;
+			proxy_connect_timeout 30s;
```
helgi: This is configurable in 1.x - might as well keep parity here

helgi: Just noticed it isn't for the k8s side but is for any other situation - I'm kinda confused why that is. Maybe someone else knows the legacy reason for that

nginx/config.go
```diff
@@ -65,8 +70,16 @@ http {
...
 			proxy_send_timeout {{ $routerConfig.DefaultTimeout }}s;
 			proxy_read_timeout {{ $routerConfig.DefaultTimeout }}s;
```
helgi: These two are configurable independently in 1.x when not on k8s but are exactly like this under k8s. Curious.

krancour: Interesting... if that is indeed a mistake, it should be fixed in a separate PR.  This one doesn't introduce it.  I will look into it further.

helgi:
All looks good from my side of it

nginx/config.go
krancour: Ah... disregarding this due to offline convo.

PR #55, opened 2015-12-14
docs(readme): remove httpd url from example

bacongobbler:
this example no longer exists


bacongobbler:
nvm, didn't see the annotation in the apache example :)

PR #57, opened 2015-12-14
feat(tests): add functional test

bacongobbler:
This PR introduces a concept called "kubeless" mode which is a workaround for running the router outside of kubernetes for functional testing purposes. It disables the kubernetes polling logic by simply blocking forever. This gives us a way to test on platforms like Travis or locally with no kubernetes cluster to test that we can run nginx.


router.go
```diff
@@ -39,3 +59,8 @@ func main() {
 		known = routerConfig
 	}
 }
+
+func forever() {
+	for {
+	}
```
arschles: I believe this is gonna spin on a core, not yielding to any other goroutines. might be a good idea to call [`runtime.Gosched()`](https://godoc.org/runtime#Gosched) inside the loop so others can do their work

router.go
```diff
@@ -10,8 +11,27 @@ import (
...
+
 func main() {
+	flag.Parse()
 	nginx.Start()
+	if kubeless {
+		// intentional loop; run forever
+		log.Println("running in kubeless mode")
+		done := make(chan bool)
+		go forever()
+		<-done
```
arschles: you can replace this and the above `make(chan...)` with `select{}`

router.go
```diff
@@ -10,8 +11,27 @@ import (
 	"k8s.io/kubernetes/pkg/util"
 )
 
+// Do we want to run nginx without kubernetes driving configuration?
+var kubeless bool
+
+func init() {
+	const (
```
technosophos: Is there a reason to put the consts inside of init?

bacongobbler: not really, just followed the [Go example for pkg flag](https://golang.org/pkg/flag/#example_) which used this convention so I went with it.

arschles: no technical harm in doing it, recommend we do this a different way and put the constants package-global and the flag handling in the main.

```go
func main() {
  kubeless := flag.Bool("kubeless", defaultKubeless, kubelessUsage)
  if *kubeless {
    ...
  }
  ...
```

krancour:
I'm almost wondering what the point of bringing this up without kube could be.  It would start nginx with absolute minimal configuration that's pretty much guaranteed to work.  What issues will we catch this way?

bacongobbler:
The only issues we will be able to catch is with spawning nginx. Perhaps the better approach is a mock configurator client to the router that will accept modification through an API a la https://github.com/progrium/nginx-appliance and https://github.com/progrium/configurator.

krancour:
I'm thinking maybe we should refactor the model a bit so that instead of passing a reference to the API to build the model, the API is used to fetch raw `metadata.annotations[deis.io/routerConfig]` and _that_ is used to generate the model.  Then "kubeless" mode in functional tests (and unit tests, for that matter) could utilize hard-coded dummy annotations.

PR #60, opened 2015-12-17
chore(glide.lock): add a glide lock file

technosophos:


arschles:
I believe this will need an upgrade the `quay.io/deis/go-dev:0.2.0` image to `latest` [in the Makefile](https://github.com/deis/router/blob/master/Makefile#L11), so that we get glide version>=0.8.0. It's currently at 0.7.2:

```console
ENG000656:workflow aaronschlesinger$ docker pull quay.io/deis/go-dev:0.2.0
0.2.0: Pulling from deis/go-dev
Digest: sha256:fa0c0251b3e9d51bccedc9413e1208c59d5635cdde9664658a0a09e0b27ed1f2
Status: Image is up to date for quay.io/deis/go-dev:0.2.0
ENG000656:workflow aaronschlesinger$ docker run --rm quay.io/deis/go-dev:0.2.0 glide --version
glide version 0.7.2
```

technosophos:
@arschles So I should use `:latest` instead of `:0.3.0`?

arschles:
Oh, didn't notice that one! I see that it has glide 0.8.1 so yes, that looks like the better choice

PR #67, opened 2015-12-28
feat(router): make platform domain optional

bacongobbler:
When a domain is not present in the routerConfig, we revert back to
a wildcard platform domain so users can make the platform domain
anything they wish.

Once this is merged, we can then remove the `domain` argument from the deis chart and move the router configuration in the docs somewhere else. This makes us compatible with the v1 router's routing behaviour.

### DIFF
--- a/model/model.go
+++ b/model/model.go
@@ -225,10 +225,12 @@ func buildAppConfig(kubeClient *client.Client, service api.Service, routerConfig
 	if err != nil {
 		return nil, err
 	}
-	if routerConfig.Domain != "" {
-		for i, domain := range appConfig.Domains {
-			if !strings.Contains(domain, ".") {
+	for i, domain := range appConfig.Domains {
+		if !strings.Contains(domain, ".") {
+			if routerConfig.Domain != "" {
 				appConfig.Domains[i] = fmt.Sprintf("%s.%s", domain, routerConfig.Domain)
+			} else {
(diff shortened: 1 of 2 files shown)


model/model.go
```diff
@@ -225,10 +225,12 @@ func buildAppConfig(kubeClient *client.Client, service api.Service, routerConfig
...
 	}
-	if routerConfig.Domain != "" {
-		for i, domain := range appConfig.Domains {
-			if !strings.Contains(domain, ".") {
+	for i, domain := range appConfig.Domains {
+		if !strings.Contains(domain, ".") {
+			if routerConfig.Domain != "" {
 				appConfig.Domains[i] = fmt.Sprintf("%s.%s", domain, routerConfig.Domain)
+			} else {
+				appConfig.Domains[i] = fmt.Sprintf("~^%s\\.(?<domain>.+)$", domain)
```
krancour: I don't think it's a good idea to have put nginx-specific configuration in the model.  Yes... several model attributes are ostensibly named after nginx directives... and there's an open issue for fixing that, but in this case, we've inserted something nginx-specific as a _value_ in the model.  This will complicate the process of swapping nginx for something else in the future (which I think is a realistic possibility).

krancour: But don't worry about this right now.  I have a fix for it, but it's backed up behind some doc changes that are more pressing.

PR #68, opened 2015-12-29
router(model): Clean model WRT to nginx-specific domain configuration

krancour:
This partly undoes #67.  It retains the functionality that was introduced there, but achieves it without the possibility of putting nginx-specific regex patterns in the model.  By not polluting the model we increase the ease with which we could swap nginx for something else in the future.


bacongobbler:
Seems fair to me. +1

PR #72, opened 2015-12-30
Do not use JSON for router configuration

### LINKED ISSUE #63: Can we not use JSON in annotations for manually configured things?
Currently we have to manually edit annotations, but annotations are in YAML in JSON even though they are just simple key/value lists:

```
⇒  kdy get rc deis-router
apiVersion: v1
kind: ReplicationController
metadata:
  annotations:
    chart.helm.sh/description: For testing only!
    chart.helm.sh/file: /Users/mattbutcher/Code/helm_home/workspace/charts/deis/manifests/deis-router-rc.yaml
    chart.helm.sh/name: deis
    chart.helm.sh/version: 2.0.0-pre-alpha
    deis.io/routerConfig: |
      {
        "domain": "10.254.1.3.xip.io",
        "useProxyProtocol": false
      }
```

Is there a reason why we can't simply make these separate annotations?

They are error prone to edit, require additional work to parse, and require additional tooling for things like Helm.

krancour:
Fixes #63 

~~This is WIP.  There are more commits to come as part of this PR.~~

### DIFF
--- a/model/model.go
+++ b/model/model.go
@@ -1,11 +1,11 @@
 package model
 
 import (
-	"encoding/json"
 	"fmt"
 	"strings"
 
 	"github.com/deis/router/utils"
+	modelerUtility "github.com/deis/router/utils/modeler"
 	"k8s.io/kubernetes/pkg/api"
 	"k8s.io/kubernetes/pkg/api/errors"
 	client "k8s.io/kubernetes/pkg/client/unversioned"
@@ -14,18 +14,18 @@ import (
 
 // RouterConfig is the primary type used to encapsulate all router configuration.
 type RouterConfig struct {
-	WorkerProcesses          string      `json:"workerProcesses"`
-	MaxWorkerConnections     int         `json:"maxWorkerConnections"`
-	DefaultTimeout           int         `json:"defaultTimeout"`
-	ServerNameHashMaxSize    int         `json:"serverNameHashMaxSize"`
-	ServerNameHashBucketSize int         `json:"serverNameHashBucketSize"`
-	GzipConfig               *GzipConfig `json:"gzipConfig"`
-	BodySize                 int         `json:"bodySize"`
-	ProxyRealIPCIDR          string      `json:"proxyRealIpCidr"`
-	ErrorLogLevel            string      `json:"errorLogLevel"`
-	Domain                   string      `json:"domain"`
-	UseProxyProtocol         bool        `json:"useProxyProtocol"`
-	EnforceWhitelists        bool        `json:"enforceWhitelists"`
+	WorkerProcesses          string      `router:"workerProcesses"`
+	MaxWorkerConnections     int         `router:"maxWorkerConnections"`
+	DefaultTimeout           int         `router:"defaultTimeout"`
+	ServerNameHashMaxSize    int         `router:"serverNameHashMaxSize"`
+	ServerNameHashBucketSize int         `router:"serverNameHashBucketSize"`
+	GzipConfig               *GzipConfig `router:"gzipConfig"`
+	BodySize                 int         `router:"bodySize"`
+	ProxyRealIPCIDR          string      `router:"proxyRealIpCidr"`
+	ErrorLogLevel            string      `router:"errorLogLevel"`
+	Domain                   string      `router:"domain"`
+	UseProxyProtocol         bool        `router:"useProxyProtocol"`
+	EnforceWhitelists        bool        `router:"enforceWhitelists"`
 	AppConfigs               []*AppConfig
 	BuilderConfig            *BuilderConfig
 }
@@ -48,13 +48,13 @@ func newRouterConfig() *RouterConfig {
 
 // GzipConfig encapsulates gzip configuration.
 type GzipConfig struct {
-	CompLevel   int    `json:"compLevel"`
-	Disable     string `json:"disable"`
-	HTTPVersion string `json:"httpVersion"`
-	MinLength   int    `json:"minLength"`
-	Proxied     string `json:"proxied"`
-	Types       string `json:"types"`
-	Vary        string `json:"vary"`
+	CompLevel   int    `router:"compLevel"`
+	Disable     string `router:"disable"`
+	HTTPVersion string `router:"httpVersion"`
+	MinLength   int    `router:"minLength"`
+	Proxied     string `router:"proxied"`
+	Types       string `router:"types"`
+	Vary        string `router:"vary"`
 }
 
 func newGzipConfig() *GzipConfig {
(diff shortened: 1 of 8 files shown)


krancour:
Still need to test this against deis/workflow#197

krancour:
Ok.  Took some doing because of unrelated issues I ran into, but this works fine in conjunction with deis/workflow#197.

krancour:
Requesting one of the reviews to be from @helgi since he's well familiar with what I'm out to achieve with this PR.

model/model.go
```diff
@@ -14,18 +14,18 @@ import (
...
-	ServerNameHashMaxSize    int         `json:"serverNameHashMaxSize"`
-	ServerNameHashBucketSize int         `json:"serverNameHashBucketSize"`
-	GzipConfig               *GzipConfig `json:"gzipConfig"`
-	BodySize                 int         `json:"bodySize"`
-	ProxyRealIPCIDR          string      `json:"proxyRealIpCidr"`
-	ErrorLogLevel            string      `json:"errorLogLevel"`
-	Domain                   string      `json:"domain"`
-	UseProxyProtocol         bool        `json:"useProxyProtocol"`
-	EnforceWhitelists        bool        `json:"enforceWhitelists"`
+	WorkerProcesses          string      `router:"workerProcesses"`
```
arschles: Can you make this tag `deis-router` instead of just `router`? My hope is that it'll reduce the probability for collisions later

arschles: same with the below tags

krancour: I'm not clear on what collisions you think might occur.  Can you elaborate, please?  From my point of view, concerns from other components aren't going to magically affect the router's own internal model and have conflicts with how it has chosen to annotate struct fields.

utils/modeler/modeler.go
```diff
@@ -0,0 +1,107 @@
...
+// MapToModel populates the provided model with values from the provided map.
+func (m *Modeler) MapToModel(data map[string]string, out interface{}) error {
+	rv := reflect.ValueOf(out)
+	return m.mapToModel(data, "", rv)
+}
+
+func (m *Modeler) mapToModel(data map[string]string, context string, rv reflect.Value) error {
+	// If rv is invalid (represents a nil litersl), we cannot proceed.
+	if rv.Kind() == reflect.Invalid {
+		return fmt.Errorf("Cannot populate a nil from the map.")
```
arschles: Can you make this and the below errors exported, so that you can document some of the expected errors that will be returned by the above `MapToModel` func? The errors with variable information in them can be `struct`s that adhere to [the `error` interface](https://godoc.org/builtin#error)

model/model.go
krancour: It should be noted that `router` in the struct tags is in no way tied to the annotation keys in the manifests.  And the latter are namespaces nicely like `router.deis.io/whatever`

utils/modeler/modeler.go
```diff
@@ -0,0 +1,107 @@
...
+	rt := elem.Type()
+	for i := 0; i < rt.NumField(); i++ {
+		rf := rt.Field(i)
+		tagValue := rf.Tag.Get(m.tag)
+		// If no tag value is found...
+		if tagValue == "" {
+			// Just move to the next field.
+			continue
+		}
+		if rf.Type.Kind() == reflect.Ptr || rf.Type.Kind() == reflect.Struct {
```
arschles: If you recurse with a pointer to a non-struct or a just a non-pointer but a struct, won't the recursion fail? Is that ok?

utils/modeler/modeler.go
krancour: Sure.

utils/modeler/modeler.go
krancour: It would fail, but not silently.  All the conditions that need to be met for handing the "top level" struct  pointer need to be satisfied when handling anything we recurse into, else an error is returned.  I can export and better document the errors as you've suggested, but I think that this is the right behavior.  We really can't do anything with anything that isn't a pointer to a struct.

model/model.go
arschles: Just talking about the struct tags, and it's mostly a paranoid thing from me. If other code decides to use a `router` struct tag, there's the conflict. And since this is an exported type, it could be used elsewhere as well.

However, given this is a router project, you're right that it's probably unlikely that there'll be a collision, so I wouldn't say it's blocking for this PR.

utils/modeler/modeler.go
arschles: Sounds good. So, since you know it'll fail before you recurse, why recurse at all? You could change the `||` to an `&&` and avoid the unnecessary recursion.

krancour: `&&` wouldn't work here.  It `rf.Type.Kind` cannot be both a `reflect.Ptr` _and_ a `reflect.Struct`, but this is a digression.

The point of this was to detect what is an obvious _attempt_ to nest models that should be populated from a map, then allow the checks for various conditions at the start of the method (when it is recursed into) to handle failure conditions without needing to repeat those checks in-line before choosing to recurse.  It's cleaner code at the expense of one additional function call.

arschles: makes sense. sounds good

arschles:
@krancour code LGTM, I haven't tested in a cluster though. Let me know if you'd like me to

krancour:
@arschles fyi... I am still working on exporting error types as you've suggested.  Doing so actually is allowing me to make a number of optimizations to the tests as well.  Stay tuned.

arschles:
sounds good @krancour. let me know if you'd like more reviews after your changes!

krancour:
@arschles error types are exported and docstring'ed now.

utils/modeler/modeler.go
```diff
@@ -0,0 +1,107 @@
...
+	tag    string
+}
+
+// NewModeler returns a pointer to a new Modeler, confgiured with the provided map key prefix and
+// struct tag key.
+func NewModeler(prefix string, tag string) *Modeler {
+	return &Modeler{prefix: prefix, tag: tag}
+}
+
+// MapToModel populates the provided model with values from the provided map.
```
arschles: would be nice to have some docs here on some of the errors to expect. not blocking for this PR, however

utils/modeler/errors.go
```diff
@@ -0,0 +1,61 @@
...
+	return fmt.Sprintf("Cannot populate non-pointer type %s from the map.", e.tipe)
+}
+
+// NilModelError represents a failed attempt to populate a "model" because a nil pointer was
+// passed to the modeler instead of a struct pointer.
+type NilModelError struct {
+	tipe reflect.Type
+}
+
+func newNilModelError(tipe reflect.Type) NilModelError {
```
arschles: since `tipe` is non-exported, this `func` should be exported

arschles: same with the below creator funcs

arschles:
@krancour awesome, thanks. made a few more comments on the new code

utils/modeler/modeler.go
krancour: I will come back to that if/when we extract this "modeler" from the router.

arschles: sounds good

utils/modeler/errors.go
krancour: What if my intention is that nothing outside the modeler package is creating errors or these types?  Exporting them is largely a courtesy because working with unexported types can be a pain, but I try to remain defensive against instantiation of any type by anyone/anything that shouldn't be doing so.

arschles: ok, that's fair - thanks for explaining. I'm good with it as-is then

PR #74, opened 2016-01-05
Remove unused config and modules

krancour:
This PR removes the `vhost_traffic_status_zone` directive.  That is part of the "virtual host traffic status module," which was only ever used for the sake of a status page that's been consciously omitted from v2 for security reasons.

With that directive gone, there are a total of 4 unused Nginx modules that this PR also removes.

### DIFF
--- a/nginx/config.go
+++ b/nginx/config.go
@@ -21,8 +21,6 @@ events {
 
 http {
 	# basic settings
-	vhost_traffic_status_zone;
-
 	sendfile on;
 	tcp_nopush on;
 	tcp_nodelay on;
--- a/rootfs/bin/build
+++ b/rootfs/bin/build
@@ -3,10 +3,6 @@
 set -eof pipefail
 
 export NGINX_VERSION=1.9.6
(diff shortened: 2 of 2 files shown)


helgi:
Custom modules and configuration around them will be something certain users will want, would be good to have a stub ticket to look into that down the line

krancour:
We never had a mechanism for arbitrary extension of the router via user-selected modules in v1.x.  The answer to anyone wanting to do that would have been to fork the router, customize, and build their own image.  I don't think that changes, does it?

helgi:
We'll at least need a documentation story around it to explain all of that, if that's not there already

Nginx 1.10 will have dynamic static object loading, as in pluggable extension system that does not require recompiling nginx. While not directly applicable in a container world then I do think extensions / modules are going to become much more accessible and used in not too distant of a future :)

krancour:
In 1.x, we can a doc section specifically about customizing each component.  I think you're right that we need something along those lines included in the new docs.

PR #75, opened 2016-01-06
feat(router): add initial ssl support

krancour:
This is the initial pass at supporting a platform-wide wildcard cert and SSL termination at the router.  It's the basic functionality.  Configurable options will come later.

Note this is a _tiny_ bit unconventional in terms of how it does things.  The cert gets managed as a k8s secret, but that does _not_ need to be mounted to the router pod(s)-- which is the unusual part.  The reason for this is pretty defensible, I think.  As I continue adding SSL support, the router will _programmatically_ obtain cert secrets for any apps having their own certs for their custom domains.  As such, for the sake of consistency, it seemed appropriate that router should obtain its own platform-wide wildcard cert using the same tactic.  It has the added benefit that Nginx only has to be reloaded when the router's cert secret is added, updated, or deleted.  The router experiences no downtime-- i.e. the pod does not need to be restarted.


krancour:
@carmstrong nevermind... answered my own question.  This linode userdata, which had looked woefully incomplete to me, gets _merged_ with the `contrib/coreos/user-data` to address linode-specific concerns.

https://github.com/deis/deis/blob/master/contrib/linode/create-linode-user-data.py#L3

PR #78, opened 2016-01-08
feat(router): add ssl support for custom domains

krancour:
This is a follow up to #75 to add SSL support for "custom domains".

This essentially generalizes some of what was done in #75 to allow certificates born by k8s secrets to be obtained and used to secure virtual hosts for any fully-qualified domain names associated with a routable service.

### DIFF
--- a/README.md
+++ b/README.md
@@ -257,15 +257,43 @@ metadata:
 
 ### SSL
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -283,12 +311,10 @@ kind: Secret
 metadata:
   name: deis-router-cert
   namespace: deis
-  labels:
-    heritage: deis
```
helgi: intentionally removed?

krancour: Yes.  Intentionally removed it from the doc because I think it's not relevant here.  This section of the docs would be most applicable to someone endeavoring to use the router in some other context than the Deis PasS, so it seemed an awkward thing to list Deis as the manifest's heritage.

PR #79, opened 2016-01-08
Edit code and docs for clarity

krancour:
This is a bundle of minor code and doc edits intended to make router a bit easier to understand on both fronts.

I've done away with the concept of "platform domain."  It's technically still there, but the old moniker didn't make much sense.  On account of the fact that the router can be used without the rest of Deis, "what platform?" might, at times, have been a relevant question.  This has been replaced with "default domain."

Similarly, I've done away with the concept of "custom domains."  _What does that even mean?_  From a deis/workflow perspective, it makes a lot of sense, but once again, from the router's perspective, loosely coupled to workflow as it is, this isn't an especially meaningful concept.

What it all boils down to, quite simply, is that for each routable service, router looks at a list of "domains."  Those might be fully-qualified domain names (which we would have called "custom" before) or simple names not containing dots which are subdomains of the "default domain" (which was "platform domain" before).

IMO, the greater precision of language introduced by this PR makes router a bit more grokable.

WIP only because this is blocked by #78, which should be merged first.


krancour:
Needs a rebase to master after #78 and _before_ this can be considered reviewable.

helgi:
I'm okay with the change but I would personally find it clearer if we didn't use `default` and instead kept referring to it as just `domain` as I find `deis-router-domain` (as a hierarchy) to be clear, and the cert on disk would be `router.cert` and so on, but do not have enough of an opinion to block this :)

PR #80, opened 2016-01-08
feat(manifests): Add liveness and readiness probes

### LINKED ISSUE #77: Add liveness and readiness probes in manifests
As in deis/charts#60

krancour:
Fixes #77

### DIFF
--- a/manifests/deis-router-rc.yaml
+++ b/manifests/deis-router-rc.yaml
@@ -32,3 +32,15 @@ spec:
           hostPort: 2222
         - containerPort: 9090
           hostPort: 9090
+        livenessProbe:
+          httpGet:
+            path: /healthz
+            port: 9090
+          initialDelaySeconds: 1
+          timeoutSeconds: 1
+        readinessProbe:
+          httpGet:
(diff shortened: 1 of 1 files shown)


manifests/deis-router-rc.yaml
```diff
@@ -32,3 +32,15 @@ spec:
...
+          httpGet:
+            path: /healthz
+            port: 9090
+          initialDelaySeconds: 1
+          timeoutSeconds: 1
+        readinessProbe:
+          httpGet:
+            path: /healthz
+            port: 9090
+          initialDelaySeconds: 1
```
helgi: Is that enough wait time? I had to bump delay to 10 on a normal go app to just get hello world but I was going from `0` :P

krancour: I _think_.  In my experience, router is ready in a lot faster than 30 seconds.  I honestly took the 30 seconds from workflow's probes and I _know_ this starts faster than workflow.  I will be happy to increase these values over time if they turn out to be too small.

PR #86, opened 2016-01-13
chore(router): Make optional gzip config a little cleaner

krancour:
The unnecessary verbosity of "gzipConfig" in the configuration was getting to me.  This PR exchanges it for just "gzip".


helgi:
very nice

PR #87, opened 2016-01-13
fix(router): Make gzip disableable again

### LINKED ISSUE #85: There is currently no way to diasble gzip compression
When we were using JSON embedded in router rc annotations to configure the router, one could disable gzip compression completely by nilling out the `gzipConfig` object in that JSON.

Since #72 we are no longer using JSON for router configuration and the ability to disable gzip compression was accidentally lost in the process.  It should be restored.

krancour:
Fixes #85


helgi:
I do not love the need for `enabled` but that's the nature of the beast I guess :(

krancour:
> I do not love the need for enabled

Nor do I, but it's the price for having ditched JSON-based config.

krancour:
At least it's defaulted to true, so in the majority of cases, no one will ever touch it or even care that this option exists.

Issue #36, opened 2015-12-09
Add support for "configuration extensions"


krancour:
In the back of my head, I want to make alternate implementations of router possible, so all along I've intended to avoid exposing configuration options that are nginx-specific.  This is easier said than done because, frankly, even with more generic names, the bulk of the router configuration _is_ nginx-specific.  Getting creative in coming up with more generic names for things only ends up obfuscating the function of each option.

To work around this, I propose adding support for implementation-specific "extensions" to the router configuration.  So, for instance, anything that is undeniably nginx specific and meaningless in the context of another hypothetical implementation (such as HAProxy), might be moved into an `"extensions": { "nginx": { ... } }` section of the configuration.

krancour:
Just capturing some further thoughts on this before I lose them...

It goes without saying that alternative implementations of the router component would _naturally_ support _different_ configuration options that vary with the capabilities, limitations, and native dialect of the underlying tech.  Options for Nginx vs HAProxy will be different and there's not much that can be done to smooth that over.  But I think that doesn't matter...

The thing that matters is that the "contract" between deis-workflow (or alternative PaaS API/controller) and the router is always honored.  When it comes down to it, there are a scant few configuration options that are affected by that and they revolve around the attributes of the `AppConfig` type.

The bottom line is that I _think_ this issue can probably be adequately resolved through a reorganization of the documentation to separate the contractual / non-negotiable aspects of the configuration from the implementation-specific aspects.

krancour:
Replaced by #111

PR #92, opened 2016-01-15
feat(router): Implement various ssl-related options

### LINKED ISSUE #89: Add support for additional SSL options
The following should be configurable as in v1.x:

* sslCiphers
* sslDhparam
* sslProtocols
* sslSessionCache
* sslSessionTickets
* sslSessionTimeout
* sslBufferSize

krancour:
Fixes #89 

~~This PR is the final bit to establish router feature parity with v1.x (minus the naxsi firewall) with v1.x.~~  This allows various SSL-related options to be fine-tuned.

Some things to note:

1.  Although this provides feature parity, defaults for some options have changed in order to earn a better grade from [Qualys SSL Labs](https://www.ssllabs.com/ssltest/analyze.html).  Just adding an appropriate cert to an app (or a wildcard cert to the platform), you can earn an A.  With some minimal tuning, an A+ is possible.

2.  There is some minor rework of things committed in #84.  The option to enforce the use of HTTPS and HSTS config options introduced there seemed like they should be pulled under the new `SSLConfig` object introduced by this PR.

3.  I intend to take a pass through _all_ configuration options (these an others) to see in what ways we can clean them up-- make them more natural to use, possibly add validation of option values, and make them less nginx-specific where possible (and implement "nginx" extensions where not possible; see #36).

### DIFF
--- a/model/model.go
+++ b/model/model.go
@@ -27,8 +27,7 @@ type RouterConfig struct {
 	DefaultDomain            string      `router:"defaultDomain"`
 	UseProxyProtocol         bool        `router:"useProxyProtocol"`
 	EnforceWhitelists        bool        `router:"enforceWhitelists"`
-	EnforceHTTPS             bool        `router:"enforceHttps"`
-	HSTSConfig               *HSTSConfig `router:"hsts"`
+	SSLConfig                *SSLConfig  `router:"ssl"`
 	AppConfigs               []*AppConfig
 	BuilderConfig            *BuilderConfig
 	DefaultCertificate       *Certificate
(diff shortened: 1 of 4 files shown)


model/model.go
```diff
@@ -133,7 +155,7 @@ type HSTSConfig struct {
 func newHSTSConfig() *HSTSConfig {
 	return &HSTSConfig{
 		Enabled:           false,
-		MaxAge:            10886400,
+		MaxAge:            15552000,
```
mboersma: There should be a trailing comment to humanize this number:
```go
MaxAge:    15552000,  // 180 days
```

krancour: Agreed.  Thank you.

PR #104, opened 2016-02-03
Config validation

### LINKED ISSUE #96: Allow validation of configuration values
Many configuration values one might set via annotations ultimately wind up in a dynamically generated `nginx.conf`.  It might be nice to detect invalid values for each field, log a warning, and fallback to a default value.  This way, Nginx itself never receives invalid configuration and logs will be more descriptive in terms of what configuration errors might exist.

krancour:
Fixes #96

This PR attempts to validate configuration values sourced from k8s annotations _before_ they make their way into dynamically generated Nginx configuration.

Before this PR, bad configuration values may have led to unfriendly error messages like this:

```
2016/02/02 23:59:20 [emerg] 14#0: "worker_processes" directive invalid value in /opt/nginx/conf/nginx.conf:4
nginx: [emerg] "worker_processes" directive invalid value in /opt/nginx/conf/nginx.conf:4
```

At minimum, the above requires anyone troubleshooting to `kubectl exec` to a router pod to inspect the resulting invalid Nginx configuration.  At worst, they may need to look at router code to get a better handle on what went wrong.

After this PR, however, declarative constraints (simple regular expressions) are used by the map -> model utility to recognize bad configuration up front.  This allows the router to pinpoint which k8s annotation is the source of the problem.  It also allows Nginx to continue functioning using default configuration in place of the bad value.

For example:

```
2016/02/02 23:51:39 WARNING: Field "router.deis.io/workerProcesses" value "otto" does not satisfy constraint /^(auto|[1-9]\d*)$/ -- skipping this field and using default value "auto".
```

### DIFF
--- a/model/model.go
+++ b/model/model.go
@@ -15,7 +15,7 @@ import (
 
 const (
 	modelerKeyPrefix     string = "router.deis.io"
-	modelerFieldTag      string = "router"
+	modelerFieldTag      string = "key"
 	modelerConstraintTag string = "constraint"
 )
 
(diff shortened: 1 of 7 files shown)


krancour:
@mboersma thanks for the LGTM.  Note that I've just amended with updated documentation.  I'll let you decide whether that warrants a second pass or not.

mboersma:
> I'll let you decide whether that warrants a second pass or not.

Nice--I will re-review.

technosophos:
I'm reviewing this now.

model/model.go
```diff
@@ -13,21 +13,32 @@ import (
...
-	ServerNameHashBucketSize int         `router:"serverNameHashBucketSize"`
-	GzipConfig               *GzipConfig `router:"gzip"`
-	BodySize                 int         `router:"bodySize"`
-	ProxyRealIPCIDR          string      `router:"proxyRealIpCidr"`
-	ErrorLogLevel            string      `router:"errorLogLevel"`
-	DefaultDomain            string      `router:"defaultDomain"`
-	UseProxyProtocol         bool        `router:"useProxyProtocol"`
-	EnforceWhitelists        bool        `router:"enforceWhitelists"`
-	SSLConfig                *SSLConfig  `router:"ssl"`
+	WorkerProcesses          string      `key:"workerProcesses" constraint:"^(auto|[1-9]\\d*)$"`
```
technosophos: I am a HUGE fan of doing constraints this way.

utils/modeler/modeler.go
```diff
@@ -59,9 +68,9 @@ func (m *Modeler) mapToModel(data map[string]string, context string, rv reflect.
 			// We're nested... use some recursion...
```
technosophos: If this is a `*int` or something, that is going to fail, no? This seems to assume that if it's a PTR, it's pointing to a struct, map, or slice. Maybe not a big deal here, though.

technosophos:
This is excellent. Nothing blocking on here. I have a minor concern with a little of the reflection code, but I don't see it as being a blocker.

utils/modeler/modeler.go
krancour: This _might_ be a problem, but since it's not introduced by this PR, I'm going to shamelessly kick that can down the road.

krancour:
Thanks for the reviews @mboersma and @technosophos.  I know this was bigger than the average PR.

utils/modeler/modeler.go
```diff
@@ -70,13 +79,30 @@ func (m *Modeler) mapToModel(data map[string]string, context string, rv reflect.
...
+				key = fmt.Sprintf("%s%s", prefix, fieldTagValue)
 			} else {
-				key = fmt.Sprintf("%s/%s.%s", m.prefix, context, tagValue)
+				key = fmt.Sprintf("%s%s.%s", prefix, context, fieldTagValue)
 			}
 			stringVal, ok := data[key]
 			if ok {
+				constraintTagValue := rf.Tag.Get(m.constraintTag)
+				if constraintTagValue != "" {
+					constraint := regexp.MustCompile(constraintTagValue)
```
arschles: Sorry I'm late to the party. This could panic, so maybe better to just call [`Compile`](https://godoc.org/regexp#Compile) and check for the error?

technosophos: Good catch. This is an odd case where while all the regexps are known at build time, they may not get evaluated until an arbitrary point in runtime, so we probably shouldn't use a Must method.

Alternately, we could precompile and cache all of the regexps in an init() and use the Must there. Then at runtime, you'd retrieve the cached copy and use it.

krancour: I do have tests for each constraint.  Those would catch any invalid regular expressions.

PR #107, opened 2016-02-03
ref(router): Add namespace to the routable label

### LINKED ISSUE #105: Routable label should be "namespaced"
Currently, the router finds "routable" apps by querying the k8s API for services with the label `routable: "true"`.

Ideally, these labels should probably change to `router.deis.io/routable: "true"`

cc @helgi: Workflow would need a complementary change.

krancour:
Fixes #105 

cc @helgi: this works fine as it turns out.

### DIFF
--- a/model/model.go
+++ b/model/model.go
@@ -14,15 +14,24 @@ import (
 )
 
 const (
-	modelerKeyPrefix     string = "router.deis.io"
+	prefix               string = "router.deis.io"
 	modelerFieldTag      string = "key"
 	modelerConstraintTag string = "constraint"
 )
 
 var (
-	namespace = utils.GetOpt("POD_NAMESPACE", "default")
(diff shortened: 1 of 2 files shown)


model/model.go
```diff
@@ -14,16 +14,26 @@ import (
 )
 
 const (
+	labelPrefix          string = "router.deis.io"
 	modelerKeyPrefix     string = "router.deis.io"
```
bacongobbler: we should just merge these both and call it `prefix`

krancour: Yep.  I'll fix that.  My initial thinking was that I wanted the variables named for their purpose, _but_ it's not exactly a coincidence that their values happen to be the same, so I'll merge them.

krancour:
@bacongobbler your feedback has been addressed.

PR #108, opened 2016-02-04
fix(router): Correct and improve whitelist behavior

### LINKED ISSUE #97: Fix the enforceWhitelists option; add support for default whitelist
Remove the enforceWhitelists option and simply enforce a whitelist for each app if one is defined, and otherwise, behave permissively.

krancour:
Fixes #97


krancour:
cc @helgi

PR #117, opened 2016-02-12
doc(router): Add generic instructions for manual LB setup

### LINKED ISSUE #59: Document how to manually configure an external load balancer
This is important for the cases where k8s is running on infrastructure where the k8s "cloud provider" uses a no-op for services of `type: LoadBalancer`.

krancour:
Fixes #59

### DIFF
--- a/README.md
+++ b/README.md
@@ -385,13 +385,78 @@ When combined with a good certificate, the router's _default_ SSL options are su
 
 Earning an A+ is as easy as simply enabling HTTP Strict Transport Security (see the `router.deis.io/ssl.hsts.enabled` option), but be aware that this will implicitly trigger the `router.deis.io/ssl.enforce` option and cause your applications to permanently use HTTPS for _all_ requests.
 
-### Front-facing load balancer and idle connection timeouts
+### Front-facing load balancer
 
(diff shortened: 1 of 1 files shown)


mboersma:
Great docs, A+, would read again. I learn something every time I read your doc PRs, Kent. Thanks!

README.md
```diff
@@ -385,14 +385,79 @@ When combined with a good certificate, the router's _default_ SSL options are su
...
+With this configuration, the path a request takes from the end-user to an application pod is as follows:
+
+```
+user agent (browser) --> front-facing load balancer --> a Deis router pod --> kube-proxy on that same node --> _any_ application pod
+```
+
+##### Option 3
+
+Option 3 is similar to option 2, but does not actually utilize a load balancer at all.  Instead, a DNS A record may be created that lists the public IP addresses of _all_ Kubernetes worker nodes.  This will leverage DNS round-robining to direct requests to all nodes.  To guarantee _all_ nodes can adequately route incoming traffic, the Deis router component should be scaled out by increasing the number of replicas specified in the replication controller to match the number of worker nodes.  Anti-affinity should ensure exactly one router pod runs per worker node.
+
```
sgoings: I did not know an A record could point to multiple IPs. :+1:

krancour:
Thanks for the reviews and positive feedback, gents!

PR #119, opened 2016-02-17
chore(ci): Always run docker build during CI

### LINKED ISSUE #118: Run docker build on all PR builds
Similar to what builder does in its [.travis.yml file](https://github.com/deis/builder/blob/master/.travis.yml#L16), router should run `make ... docker-build` to ensure that the Dockerfile is correct

krancour:
Fixes #118

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -23,7 +23,7 @@ GO_PACKAGES := ${REPO_PATH} $(addprefix ${REPO_PATH}/,${GO_DIRS})
 # The following variables describe the Docker image we build and where it
 # is pushed to.
 # If DEIS_REGISTRY is not set, try to populate it from legacy DEV_REGISTRY.
-DEIS_REGISTRY ?= ${DEV_REGISTRY}/
+DEIS_REGISTRY ?= ${DEV_REGISTRY}
 IMAGE_PREFIX ?= deis
 IMAGE := ${DEIS_REGISTRY}${IMAGE_PREFIX}/${SHORT_NAME}:${VERSION}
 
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -23,7 +23,7 @@ GO_PACKAGES := ${REPO_PATH} $(addprefix ${REPO_PATH}/,${GO_DIRS})
 # The following variables describe the Docker image we build and where it
 # is pushed to.
 # If DEIS_REGISTRY is not set, try to populate it from legacy DEV_REGISTRY.
-DEIS_REGISTRY ?= ${DEV_REGISTRY}/
+DEIS_REGISTRY ?= ${DEV_REGISTRY}
```
krancour: @arschles can I beg a re-review from you?

You'd previously asked why I bothered to set the `DEIS_REGISTRY` environment variable in the `.travis.yml` and I agreed it wasn't needed.  It was... and here's why...

This line was screwing me over.  In Deis 1.x, we used `DEV_REGISTRY` and it never ended with a trailing slash as `DEIS_REGISTRY` does.  I had, once upon a time, tried to be smart about this legacy support by adding in that trailing slash, but it actually causes `DEIS_REGISTRY` to be just `/` in the case where neither of those vars is initially defined.  And that makes the `docker build` complain with the following error:

```
docker build --rm -t /deis/router:git-a6ddf34 rootfs
repository name component must match "[a-z0-9](?:-*[a-z0-9])*(?:[._][a-z0-9](?:-*[a-z0-9])*)*"
```

By _not_ adding the trailing slash in here, it will get this PR passing _and_ will improve consistency with the other components.

arschles: @krancour makes sense, and this LGTM!

PR #121, opened 2016-02-18
ref(router): Run as a non-root user

### LINKED ISSUE #120: Run container as non-root user
Right now the container is run as root.

krancour:
Fixes #120 

This wasn't easy.  Running router as a non-privileged user means router can't su to start Nginx as a second unprivileged user `nginx`.  The solution is to move _everything_-- all files and processes under a new `router` user.  This works great except for a limitation that unprivileged users cannot bind to low-numbered/privileged ports like 80 and 443.  That's where `setcap` comes in.

### DIFF
--- a/rootfs/Dockerfile
+++ b/rootfs/Dockerfile
@@ -9,26 +9,34 @@ RUN apk add --update-cache \
 	openssl \
 	pcre \
 	sudo \
+	libcap \
 	&& rm -rf /var/cache/apk/*
 
-# add nginx user
-RUN addgroup -S nginx && \
-  adduser -S -G nginx -H -h /opt/nginx -s /sbin/nologin -D nginx
+# add router user
+RUN addgroup -S router && \
+	adduser -S -G router -H -h /opt/router -D router
 
 COPY . /
 
 # compile nginx from source
 RUN build
 
-# Re-copy this file because the previous step will have overwritten it
-COPY opt/nginx/conf/nginx.conf /opt/nginx/conf/nginx.conf
+# Re-copy these files because the previous step will have overwritten them
+COPY opt/router /opt/router
 
-RUN mkdir /opt/nginx/ssl
+RUN mkdir /opt/router/ssl
 
-RUN ln -sf /dev/stdout /opt/nginx/logs/access.log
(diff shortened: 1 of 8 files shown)


mboersma:
Nicely done--`setcap` FTW. :+1:

rootfs/Dockerfile
```diff
@@ -9,26 +9,34 @@ RUN apk add --update-cache \
...
+# Re-copy these files because the previous step will have overwritten them
+COPY opt/router /opt/router
 
-RUN mkdir /opt/nginx/ssl
+RUN mkdir /opt/router/ssl
 
-RUN ln -sf /dev/stdout /opt/nginx/logs/access.log
-RUN ln -sf /dev/stderr /opt/nginx/logs/error.log
+RUN ln -sf /dev/stdout /opt/router/logs/access.log
+RUN ln -sf /dev/stderr /opt/router/logs/error.log
```
arschles: worth putting these in the same `RUN` command? same with the below 2 `RUN`s

krancour:
@arschles your feedback has been applied.

rootfs/Dockerfile
```diff
@@ -9,26 +9,34 @@ RUN apk add --update-cache \
...
-RUN ln -sf /dev/stdout /opt/nginx/logs/access.log
-RUN ln -sf /dev/stderr /opt/nginx/logs/error.log
+RUN ln -sf /dev/stdout /opt/router/logs/access.log && \
+	ln -sf /dev/stderr /opt/router/logs/error.log
 
-CMD ["/bin/router"]
+# Fix some permission since we'll be running as a non-root user.  This includes enabling
+# Nginx to always bind to low/privileged ports, even when running as a non-root user.
+RUN chown -R router:router /opt/router && \
+	setcap 'cap_net_bind_service=+ep' /opt/router/sbin/nginx
```
arschles: @krancour is it worth merging the two commands here with those in the above `RUN`? not blocking for this PR, just want your thoughts.

krancour: Can't... the permission changes on /opt/router have to occur _after_ the last copy from rootfs.

PR #124, opened 2016-02-18
fix(router): Fix log permissions for non-root user

krancour:
#120 turned out to not work nicely with Docker < 1.9.

This PR corrects that.

cc @arschles


arschles:
@krancour LGTM. is there any possible way to test the perms in CI? not blocking for this PR

krancour:
For reference, this is the bug in Docker < 1.9 that this works around:

https://github.com/deis/workflow/pull/398

PR #126, opened 2016-02-19
Rollback running as non-root user

krancour:
This rolls back running router as a non-root user since that is causing problems on gke and is thus holding up e2e testing.

cc @arschles


krancour:
This is superceded by #127.  Closing.

PR #133, opened 2016-02-28
chore(Makefile): update go-dev to 0.10.0 and compress router binary

mboersma:
```
$ docker images
REPOSITORY                     TAG       IMAGE ID       CREATED             SIZE
192.168.64.2:5000/deis/router  after     c445c3459a68   4 minutes ago       38.14 MB
192.168.64.2:5000/deis/router  before    6b62b3583f0e   11 minutes ago      64.37 MB
`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,4 @@
 manifests/deis-router-rc.tmp.yaml
-rootfs/bin/router
 rootfs/opt/router/sbin/router
+rootfs/opt/router/sbin/router.*
 vendor/
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 manifests/deis-router-rc.tmp.yaml
+rootfs/bin/router
 rootfs/opt/router/sbin/router
 vendor/
--- a/Makefile
+++ b/Makefile
@@ -10,7 +10,7 @@ REPO_PATH := github.com/deis/${SHORT_NAME}
 
 # The following variables describe the containerized development environment
 # and other build options
-DEV_ENV_IMAGE := quay.io/deis/go-dev:0.9.0
+DEV_ENV_IMAGE := quay.io/deis/go-dev:0.10.0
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

File: .gitignore [hunk 1]
```diff
@@ -1,3 +1,5 @@
 manifests/deis-router-rc.tmp.yaml
+rootfs/bin/router
```
krancour: I don't think this addition is required.  What is most likely is that you have a vestigial binary remaining in this old location from older build.  We used to put it there... but then it moved to its present location.

.gitignore
```diff
@@ -1,3 +1,5 @@
...
 rootfs/opt/router/sbin/router
+rootfs/opt/router/sbin/router.*
```
krancour: Not sure I understand this addition.

.gitignore
mboersma: You're right. I should have run `git clean`. :-)

.gitignore
mboersma: `upx` created some temp binaries apparently. I had `router.upx` and `router.000` binaries there somehow while testing this PR.

krancour: Ok.  That makes sense.

mboersma:
Updated to use [`deis/go-dev:0.9.1`](https://github.com/deis/docker-go-dev/releases/tag/0.9.1).

Makefile
```diff
@@ -50,6 +53,8 @@ bootstrap: check-docker
 build: check-docker
 	mkdir -p ${BINDIR}
 	${DEV_ENV_CMD} make binary-build
+	$(call check-static-binary,$(BINDIR)/${SHORT_NAME})
+	${DEV_ENV_CMD} ${GOUPX} ${BINDIR}/${SHORT_NAME} || exit 1
```
sgoings: Why do we need this `|| exit 1`?

mboersma:
Updated to use [`deis/go-dev:0.10.0`](https://github.com/deis/docker-go-dev/releases/tag/0.10.0).

PR #138, opened 2016-03-08
fix(ssl): Don't ever use the default/platform-wide wildcard cert for "custom" (fully qualified) domains

### LINKED ISSUE #137: Default cert should not be used for "custom" domains
This is similar to, but different from #136.  The two issues should not be confused.

Currently, if you _do_ provide a platform-wide wildcard cert, that cert gets used for securing "custom" (fully-qualified) domains associated to an app-- and that's wrong because the cert is not going to match that domain.  Every "custom" domain should only support SSL if a cert is explicitly associated to it.

krancour:
Fixes #137

### DIFF
--- a/model/model.go
+++ b/model/model.go
@@ -343,8 +343,8 @@ func buildAppConfig(kubeClient *client.Client, service api.Service, routerConfig
 	}
 	// Step through the domains, and decide which cert, if any, will be used for securing each.
 	// For each that is a FQDN, we'll look to see if a corresponding cert-bearing secret also
-	// exists.  If so, that will be used.  If a domain isn't an FQDN OR a it is, but a corresponding
-	// cert-bearing secret does not exist, we will use the default cert-- even if that is nil.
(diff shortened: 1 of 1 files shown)


arschles:
@krancour what should we do to test this? Can you include some instructions?

vdice:
From @krancour:
```
to test... add a wildcard domain to the platform.  then add a "custom" domain (fqdn) for some app, ​_without_​ also adding a cert for that custom domain.  then request it over https... the result should be a cert warning... if you ignore the warning, you should get a 404
```
I'll test this now...

vdice:
Tested after adding a wildcard platform cert for `*.foo.com` and testing https for custom `bob.foo.com` domain:
```
$ curl -kI -H "Host: bob.foo.com" https://app.130.211.185.40.xip.io
HTTP/1.1 404 Not Found
Server: nginx/1.9.6
Date: Wed, 09 Mar 2016 22:57:35 GMT
Content-Type: text/html
Content-Length: 168
Connection: keep-alive
```

PR #146, opened 2016-03-15
chore(Makefile): Enable immutable (git-based) Docker tag publishing

### LINKED ISSUE #145: Publish immutable artifact tags + canary version to Docker registries
We want to have the ability to have deis/charts reference a specific immutable image which will enable us better traceability on issues that come into various components.

What we've done with efforts in other repositories is publish a version:
- `git-[short sha]` that will never be overwritten
- `canary` that will move to reference the latest version available of a component

Examples of the easiest way to roll this change out can be seen in deis/workflow#487, deis/builder#246, and deis/workflow-manager#11

krancour:
Fixes #145

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,8 +1,10 @@
 include includes.mk
 
 SHORT_NAME := router
-GIT_SHA := $(shell git rev-parse --short HEAD)
-VERSION ?= git-${GIT_SHA}
+DEIS_REGISTRY ?= ${DEV_REGISTRY}
+IMAGE_PREFIX ?= deis
+
+include versioning.mk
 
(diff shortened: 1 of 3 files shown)


sgoings:
Thanks @krancour!

krancour:
I made some edits cause of a goof.  Should be better now.

sgoings:
Don't forget you'll also have to make an update to deis/charts like: https://github.com/deis/charts/pull/159

krancour:
Yerp.  Thanks for the reminder!

arschles:
LGTM3

PR #153, opened 2016-03-23
feat(router): Respond with 503 for upstream svcs w/ no endpoints

### LINKED ISSUE #150: Router should respond with 503 when an upstream service has no endpoints
Router used to do this, but it was judged unnecessary, so when a bug was detected in that bit of code, it was removed instead of fixed.

See #44 for decision to remove.  (Note that the code linked to in the issue is wrong, because it references lines of a file in the _master_ branch.  The contents of that file have long since changed and shifted around.)

See deis/workflow-e2e#122 for an explanation of why this feature needs to be re-introduced.

krancour:
Fixes #150 

Relates to deis/workflow-e2e#122

Necessitates the following change in [workflow-e2e](https://github.com/deis/workflow-e2e):
* [ ] https://github.com/deis/workflow-e2e/issues/127

### DIFF
--- a/model/model.go
+++ b/model/model.go
@@ -110,6 +110,7 @@ type AppConfig struct {
 	ServiceIP      string
 	CertMappings   map[string]string `key:"certificates" constraint:"(?i)^((([a-z0-9]+(-[a-z0-9]+)*)|((\\*\\.)?[a-z0-9]+(-[a-z0-9]+)*\\.)+[a-z]{2,}):([a-z0-9]+(-[a-z0-9]+)*)(\\s*,\\s*)?)+$"`
 	Certificates   map[string]*Certificate
+	Available      bool
 }
 
 func newAppConfig(routerConfig *RouterConfig) *AppConfig {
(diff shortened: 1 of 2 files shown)


arschles:
@krancour Code LGTM. Also, can you assign this a milestone?

krancour:
@arschles I can... but not clear on whether we're sneaking this into beta or not.

arschles:
Ok. My vote is for not putting it in the beta (although I very much like this change)

@slack can you weigh in here?

slack:
How does this look in e2e independent of the fixes @smothiki and @vdice are working on?

smothiki:
@krancour if this fix goes in for beta then I have to deal with a different logic for tests. Let me know if you are thinking of merging this for beta ??

krancour:
@smothiki would this change any more than the response code you look for when an app is scaled to zero?

krancour:
@slack it would fail e2e currently.

smothiki:
Im generally checking for either 503 or timeout and passing the tests for a 0 scaled applications

slack:
Ok, I think RC makes sense.

smothiki:
My changes are actually independent of this PR , so no worries

vdice:
Fixes https://github.com/deis/controller/issues/552

PR #162, opened 2016-04-13
tests(Makefile): add test-style shellcheck target

mboersma:


krancour:
imo, we'd be better off adding shellcheck to the go-dev image. It's certainly used often enough among our projects. Also, as it stands now, all of the build and tests (including existing style / format checks) are containerized.  Having Travis download shellcheck requires that particular check to happen _outside_ the container, which is an inconsistency / idiosyncrasy we could avoid by including shellcheck in the image.  Thoughts?

mboersma:
> we'd be better off adding shellcheck to the go-dev image.

Yes, that would be simpler and cleaner. Let's get deis/docker-go-dev#31 (and deis/docker-go-dev#30) merged and tag a new release, then I'll rework this and my other `shellcheck` PRs to use the go-dev image.

mboersma:
@krancour I refactored this to use `shellcheck` from docker-go-dev:0.11.0. Much better. [Travis is happy](https://travis-ci.org/deis/router/builds/122902248):
```
docker run --rm -v /home/travis/build/deis/router:/go/src/github.com/deis/router -w /go/src/github.com/deis/router quay.io/deis/go-dev:0.11.0 shellcheck _scripts/deploy.sh _scripts/generate-changelog.sh rootfs/bin/build rootfs/bin/get_src rootfs/opt/router/sbin/boot
```

I'll take this approach in the other `shellcheck` PRs I created.

Makefile
```diff
@@ -89,6 +91,7 @@ test-functional:
 	@echo no functional tests
 
 test-style: check-docker
+	${DEV_ENV_CMD} shellcheck $(SHELL_SCRIPTS)
```
krancour: @mboersma what would you think about moving the shellcheck under the `style-check` target that this tagret already runs within a container (see next line)?  This way we're slightly more efficient by only bootstrapping the container once and executing _all_ style checks in that one container.

PR #186, opened 2016-05-12
chore(rootfs/Dockerfile): DEIS_RELEASE -> WORKFLOW_RELEASE

vdice:
Update `DEIS_RELEASE` to be `WORKFLOW_RELEASE` as done similarly elsewhere (see https://github.com/deis/workflow/pull/193)


mboersma:
I was talking with @vdice about this: I'm pretty sure the entire `DEIS_RELEASE` (now `WORKFLOW_RELEASE`) environment variable in the Dockerfiles is a vestige of Deis v1 that we don't need now. I'm going to ask a couple other knowledgeable folks and if so I'll make a proposal and we can have a separate set of PRs to clean this up.

Until then, carry on! 🏇

PR #189, opened 2016-05-20
chore(Dockerfile): Refactor image to use ubuntu-slim

krancour:


krancour:
@arschles the speter thing has reared its head again and that's what's failing CI here.

Any chance you can fix me up?

smothiki:
Running a user as root or non root doesn't make much difference unless we are mounting some volumes. But all the secrets should be read only or else might mess up something

krancour:
> Running a user as root or non root doesn't make much difference unless we are mounting some volumes.

That's not entirely true. Docker containers all share the hosts kernel. afaik, it's a best practice to not run things as root in containers when you can avoid it.  As long as the implicit `USER` is root, anyone exec'ing into the container is also root and can do Bad Things™.

That being said, running router as a non-root user isn't new to this PR. We've been doing it for several months now with now ill effects.

PR #193, opened 2016-06-01
feat(ssl): Delete unused certs and dhparam

### LINKED ISSUE #144: Delete unused certs
Presently, the router doesn't delete unused certs.  They stick around on disk for the life of the container.  The reason no cleanup occurs is because a. it doesn't strictly _need_ to, and b. it's hard-ish.  Router gets a list of certs that are needed from annotations on each routable service, but it doesn't have a list of certs that are _no longer_ needed.  Determining which are no longer needed would involve comparing those extant on disk to those that are needed.

However, as @helgi has pointed out, on each config change, we could possibly consider nuking _all_ certs and writing / re-writing just those that are used.  I'm :+1: on this, although I feel it's a low priority issue.

krancour:
Closes #144 

In addition to what #144 called for, this does the same for any custom dhparam that may have been used-- deletes it if it is no longer used.

### DIFF
--- a/nginx/config.go
+++ b/nginx/config.go
@@ -4,7 +4,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"os"
-	"path"
+	"path/filepath"
 	"text/template"
 
 	"github.com/Masterminds/sprig"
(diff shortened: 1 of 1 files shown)


nginx/config.go
```diff
@@ -224,16 +224,38 @@ http {
...
+	allCertsGlob, err := filepath.Glob(filepath.Join(sslPath, "*.crt"))
+	if err != nil {
+		return err
+	}
+	allKeysGlob, err := filepath.Glob(filepath.Join(sslPath, "*.key"))
+	if err != nil {
+		return err
+	}
+	for _, cert := range allCertsGlob {
+		err = os.Remove(cert)
```
vdice: I believe this block (and similar below) can technically be reduced to
```
if err := os.Remove(cert); err != nil {
  return err
}
```

krancour: Fixing it. It's probably a matter of personal style that I rarely collapse them like that. Probably because 9 times out of 10, I'm getting multiple return values from a function call and have big plans for that other variable. But in this case, it can be done, so I will.

krancour: Fixed.

PR #194, opened 2016-06-01
chore(ssl): Remove support for old cert secret format

krancour:
#158 added support for differently formatted k8s secrets conveying SSL certs. The new format was intended to match that documented for use with ingresses [here](http://kubernetes.io/docs/user-guide/ingress/#tls). #158 was tactical, however, and retained support for the old format as well. Support for the old format should be removed (as was always intended) before we go GA.

cc @helgi

### DIFF
--- a/model/model.go
+++ b/model/model.go
@@ -394,20 +394,14 @@ func buildCertificate(certSecret *api.Secret, context string) (*Certificate, err
 	cert, ok := certSecret.Data["tls.crt"]
 	// If no cert is found in the secret, warn and return nil
 	if !ok {
-		cert, ok = certSecret.Data["cert"]
-		if !ok {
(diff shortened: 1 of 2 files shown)


model/model.go
```diff
@@ -309,7 +309,7 @@ func buildRouterConfig(rc *api.ReplicationController, platformCertSecret *api.Se
 		return nil, err
 	}
 	if platformCertSecret != nil {
-		platformCertificate, err := buildCertificate(platformCertSecret, "default")
+		platformCertificate, err := buildCertificate(platformCertSecret, "platform")
```
helgi: Should this change be intermixed with the rest?

krancour: The original text was wrong-- we stopped calling it the "default" cert a long time ago. (That actually means something else.) But, it's probable that should be a separate fix. I'll amend this and submit a separate PR for that. Thanks for keeping me honest.

krancour: Fixed.

PR #203, opened 2016-06-16
feat(model): expose any ports in the ephemeral port range

bacongobbler:
This PR exposes any ports in the ephemeral port range from the service as `http://foo.myapp.com:32768`. The problem with this approach lies in the limitation with kubernetes where you cannot specify a port range to expose in a service. This was more of a learning exercise than something we could implement, but pinging @krancour anyways to see if there's any way we could get to support apps that expose more than one port, a la https://github.com/deis/workflow/issues/324.

I'd love to support *any* port, however given that we already reserve a couple ports for other in-house applications and port 2222 is strictly dedicated to the builder via TCP, I do not see a way around this.

We do a couple of things here:

 - check if the port protocol is TCP (k8s blankets HTTP connections under TCP)
 - check if the port is either 80 (for backwards compatibility) as well as any in the ephemeral range (ports 32768-61000)
 - if yes, add it to the ServicePorts field and template it out for every domain for said app.

The other significant change which can occur outside of this PR is the change to run nginx as root. nginx has been running as root for production cases for many years without issue, because worker processes do not run as root (by default they run under the `nginx` user). This is necessary so the port logic implemented in this PR is a 1:1 mapping, rather than having to special-case proxying requests from the router's port 8080 to the application's port 80.


krancour:
This mostly looks pretty good to me. I certainly like what it is trying to accomplish. I do have feedback on a few things tho.

My biggest disagreements have mainly to do with listening on port 80 and running as root. Bear with me...

> The other significant change which can occur outside of this PR is the change to run nginx as root.

This probably isn't true. If this PR makes Nginx bind to port 80 (instead of 8080 as it was previously), to my knowledge, Nginx will not start. Port 80 is a privileged port. Binding to it would require that the transition to running as root be included in this PR.

Having said that, I strongly believe we should not do that. The transition to not running as root was deliberate and relatively difficult, though worth it, I believe.

> nginx has been running as root for production cases for many years without issue, because worker processes do not run as root (by default they run under the nginx user).

This is a valid point. However, that has generally _not_ been in containers for most of that time. Containers change the equation pretty significantly. Set aside any possibility of an Nginx vulnerability being exploited remotely by an attacker and consider instead in the normal course of operations, certain operators may exec into a container for troubleshooting purposes. If Nginx runs as root, root is also the user whose identity the operator assumes when execing into the container. As root, that operator has unrestricted access to the kernel-- which is shared with the host and with all other containers on that host. Now consider that even if authorized to exec into containers, this individual might not be authorized to remotely access the hosts (k8s worker nodes) directly. Running _anything_ as root inside a container represents a back door into the host's kernel. Yes... I know that this is only dangerous if you don't trust the people in your own org. But a. "inside jobs" do happen and b. security experts site fears of the shared kernel being exploited as their number one reservation about containers. Frankly, we should be paying more attention to not running _all_ our components as root, rather than moving this one in the opposite direction. (Also consider that everything I just said is a major consideration for large enterprises operating in regulated industries.)

Having said all _that..._ I don't think that running Nginx as a non-root user, and the inability to bind Nginx to port 80 necessarily interferes with what you are trying to achieve.

Some things to consider:

* No matter what unprivileged port(s) that processes within a container bind to, that can _always_ be compensated for with a k8s service. So, for instance, if an application, within the container listens on 5000, the _service_ can still expose port 80 and map that back to container port 5000. No client of said service is any the wiser that the application isn't actually listening on 80.
* From the router's perspective, it's that above _service_ port that is actually important.
* afaik, those service ports could be _any_ number from 1 to 65535, without any trouble.

Ultimately, as I said, I like what this PR is trying to achieve, but I think that via services, k8s offers enough flexibility that there isn't any need to make Nginx run as root or bind to privileged ports. I'm certain that can be achieved and I'd be happy to help with it.

bacongobbler:
Sure. I'll see if I can get this to work while nginx runs as non-root. After thinking about it this morning I'm sure there's a way I can work around it without having to go this route. :)

> No matter what unprivileged port(s) that processes within a container bind to, that can always be compensated for with a k8s service. So, for instance, if an application, within the container listens on 5000, the service can still expose port 80 and map that back to container port 5000. No client of said service is any the wiser that the application isn't actually listening on 80.

This is brilliant, too. Thinking more on this, we can document that any non-default routable port will be available in the ephemeral port range (for workflow), and that using the router outside of workflow would require the service to use a port in the ephemeral range, then map that back to the container's port 5000.

krancour:
> we can document that any non-default routable port will be available in the ephemeral port range (for workflow), and that using the router outside of workflow would require the service to use a port in the ephemeral range, then map that back to the container's port 5000.

That sounds right, but I just want to make sure I understand that properly. There's really no restriction on what ports a process can bind to within a container (unless they aren't root, but we have no say in whether processes within a Dockerfile or `deis pull` app run as root). And there's no restriction at all re: what port(s) a service exposes. So the restriction to ephemeral ports is imposed then by router for lack of ability to bind to privileged ports?  Am I reading that correctly?

If I have that right, also consider that we might not even need to restrict it to ephemeral ports. We can restrict it just to unprivileged ports-- which is a much broader range.

But now a couple things are popping into my head that could be bigger challenges than anything we have discussed yet...

The first, you already had mentioned:

> The problem with this approach lies in the limitation with kubernetes where you cannot specify a port range to expose in a service.

I guess it's _possible_ that we could make the controller add ports to the router's own service as needed. But that increases coupling between the controller and router.

The other issue... if an application exposes more than one port, there's a decent chance that one of those ports is _not_ for HTTP. What if it's for SMTP or FTP or some other application level protocol? The way that has to be dealt with is much different-- more similar to the `stream` that we're using for the builder. Even if we could make that work, we could end up with unresolvable conflicts. Where HTTP is concerned, it's fine for multiple apps to use the same port, because virtualhosts are also taken into account, but no such analog exists for a `stream`-- which uses raw TCP/IP.

bacongobbler:
closing as I do not think this will come to fruition. however, this is some good food for thought to help tackle deis/workflow#324.

PR #205, opened 2016-06-17
feat(*): Update nginx to 1.10.1

helgi:
http://nginx.org/en/CHANGES-1.10

Intentionally not going for the unstable branch, until features from there are needed

### DIFF
--- a/rootfs/Dockerfile
+++ b/rootfs/Dockerfile
@@ -17,12 +17,12 @@ RUN apt-get update \
 		libssl-dev \
 		libpcre3-dev \
 		make \
-	&& export NGINX_VERSION=1.9.6 VTS_VERSION=0.1.8 BUILD_PATH=/tmp/build PREFIX=/opt/router \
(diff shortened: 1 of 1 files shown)


krancour:
The router has _so many_ options that covering them all with e2e has not been a realistic possibility. I think this justifies adding the manual testing label to this.

mboersma:
This needs to update the SHA for the nginx tarball download:
```console
$ shasum -a 256 nginx-1.10.1.tar.gz 
1fd35846566485e03c0e318989561c135c598323ff349c503a6c14826487a801  nginx-1.10.1.tar.gz
```

helgi:
Thanks, we should probably move away from sha's if nginx isn't providing them

krancour:
pgp would be better.

helgi:
agreed - lets do it as a separate issue tho

mboersma:
I've been running this on my local cluster for a while, and put it through my other `boom` and e2e tests. Seems to work. 👍

PR #216, opened 2016-07-09
Enable HTTP/2 support in the router

### LINKED ISSUE #204: Add http2 listen support.

felixbuenemann:
This enables HTTP/2 over TLS in the DEIS Workflow router.

The config change is all that's needed, because NGINX is [already compiled with the http_v2 module](https://github.com/deis/router/blob/a8497aff73027d330abf39d27d92194ad93e61b9/rootfs/Dockerfile#L45) and ALPN support (provided by openssl 1.0.2+).

This fixes #204.

You can pull image [docker.io/felixbuenemann/router:canary](https://hub.docker.com/r/felixbuenemann/router/) for manual testing.

### DIFF
--- a/nginx/config.go
+++ b/nginx/config.go
@@ -115,7 +115,7 @@ http {
 	# Default server handles requests for unmapped hostnames, including healthchecks
 	server {
 		listen 8080 default_server reuseport{{ if $routerConfig.UseProxyProtocol }} proxy_protocol{{ end }};
-		listen 6443 default_server ssl{{ if $routerConfig.UseProxyProtocol }} proxy_protocol{{ end }};
+		listen 6443 default_server ssl http2{{ if $routerConfig.UseProxyProtocol }} proxy_protocol{{ end }};
 		set $app_name "router-default-vhost";
 		{{ if $routerConfig.PlatformCertificate }}
 		ssl_protocols {{ $sslConfig.Protocols }};
@@ -166,7 +166,7 @@ http {
 		set $app_name "{{ $appConfig.Name }}";
 
 		{{ if index $appConfig.Certificates $domain }}
-		listen 6443 ssl{{ if $routerConfig.UseProxyProtocol }} proxy_protocol{{ end }};
+		listen 6443 ssl http2{{ if $routerConfig.UseProxyProtocol }} proxy_protocol{{ end }};
 		ssl_protocols {{ $sslConfig.Protocols }};
 		{{ if ne $sslConfig.Ciphers "" }}ssl_ciphers {{ $sslConfig.Ciphers }};{{ end }}
 		ssl_prefer_server_ciphers on;


deis-admin:

felixbuenemann:
A simple commandline test that http2 with ALPN is working can be achieved using:

```
openssl s_client -connect deis.example.com:443 -alpn 'h2'
…
New, TLSv1/SSLv3, Cipher is ECDHE-RSA-AES256-GCM-SHA384
Server public key is 2048 bit
Secure Renegotiation IS supported
Compression: NONE
Expansion: NONE
ALPN protocol: h2
…
```

To test older NPN negotiation:
```
openssl s_client -connect deis.example.com:443 -nextprotoneg 'h2'
…
New, TLSv1/SSLv3, Cipher is ECDHE-RSA-AES256-GCM-SHA384
Server public key is 2048 bit
Secure Renegotiation IS supported
Compression: NONE
Expansion: NONE
Next protocol: (1) h2
No ALPN negotiated
…
```

arschles:
I'm ok with this change. I'd like @krancour to also take a look at this patch (if available), but I'm `LGTM1`ing it

krancour:
I'm a little late to the game here, but this does not _exclude_ the use of HTTP 1.1, correct?

helgi:
Nope

krancour:
@felixbuenemann thank you very much for this contribution!

felixbuenemann:
@krancour If you enable http2 in nginx on a `listen ssl` port, it wil use the TLS NPN and ALPN extensions for negotiating a supported protocol with the client. If the client does not support either NPN or ALPN it will simply speak HTTP/1.1, HTTP/1.0 or HTTP/0.9.

If you however put http2 on a non-ssl port in nginx, that port will *only* speak HTTP/2, so you will need a load balancer like haproxy in front of it, which handles ssl termination and protocol negotiation and sends traffic to different ports based on the negotiated protocol.

krancour:
> If you however put http2 on a non-ssl port in nginx, that port will only speak HTTP/2

I feel a little concerned by that statement. Our preference is any load balancer in front of the router is TCP-based and not HTTP-based.

felixbuenemann:
@krancour No need to be concerned.

This is fine:

```
listen 443 ssl http2;
```
In this case port 443 speaks both HTTP/1.1 and HTTP/2. If the client does support neither NPN nor ALPN it will use whatever non HTTP/2 protocol it chooses to speak. NGINX support HTTP/0.9, HTTP/1.0 and HTTP/1.1 in this case. If the client supports NPN or ALPN it will negotiate with NGINX to use either HTTP/2 (h2) or HTTP/1.1.

This is not fine:

```
listen 80 http2;
```

That would switch port 80 to *only* speak http2 and break *all* clients, because there is no way to negotiate wether a client wants to speak http or http/2.

So to get a working setup you would need something like:

```
listen 80 proxy_protocol;
listen 81 http2 proxy_protocol;
``` 

And slap a HAProxy in front of it which terminates SSL (not HTTPS), negotiates HTTP/2 over ALPN and forwards the traffic to the right port, 80 for http and 81 for http/2.

We're only using http/2 on the ssl port, so none of the above is needed.

Did I make myself sufficiently clear this time?

krancour:
> We're only using http/2 on the ssl port

That's the part I overlooked. Thanks.

codyaray:
If I understand correctly, this router terminates the HTTP/2 connection and opens an HTTP/1.1 connection to the app rather than proxy/passthrough the HTTP/2 connection. If we want to proxy the HTTP/2 all the way to the backend, can Deis route/load balance TCP connections?

bacongobbler:
@codyaray you are indeed correct, [according to the source](https://github.com/deis/router/blob/master/nginx/config.go#L221). Feel like hacking up a PR to proxy http/2 connections all the way to the backend?

bacongobbler:
Or at the very least filing a new issue to track that

felixbuenemann:
I don't think nginx can do ssl vhosts in tcp mode, let alone mixing tcp and https vhosts. The only tool I know that can route to multiple tcp backends based on the SNI header is HAProxy.

PR #218, opened 2016-07-21
fix(router): BodySize config can be "0"

olalonde:
Setting size to 0 disables checking of client request body size.

See http://nginx.org/en/docs/http/ngx_http_core_module.html#client_max_body_size


deis-admin:

Joshua-Anderson:
Looks like tests are failing because of an [assertion that needs to be updated](https://github.com/deis/router/blob/master/model/model_validation_test.go#L67)

olalonde:
Thanks, think I fixed it.

PR #222, opened 2016-08-03
feat(*): Update nginx to 1.11.2

### LINKED ISSUE #221: HTTP2 issue with many clients
Recently #216 added HTTP2 support to the router, which is awesome, however a bug exists in the HTTP2 implementation in 1.10.x which causes a bunch of clients to vomit, in my case OkHttp on Android. See more here: https://trac.nginx.org/nginx/ticket/959.

Three options exist:
1.) Upgrade to NGINX 1.11.x(mainline) in which the issue is resolved.
2.) Remove the HTTP2 support from the router until 1.12.x is released to stable.
3.) Allow HTTP2 support to be enabled/disabled via an annotation on the router.

I'm working on trying option 1 on my staging cluster to see if there are any issues with a simple version bump.

croemmich:
Fixes #221.

Note, not using 1.11.3 due to a nginx bug: https://trac.nginx.org/nginx/ticket/1032

Also bumped VTS module as it advertises compatibility. Manually verified that stats are still flowing.

### DIFF
--- a/rootfs/Dockerfile
+++ b/rootfs/Dockerfile
@@ -17,14 +17,14 @@ RUN apt-get update \
 		libssl-dev \
 		libpcre3-dev \
 		make \
-	&& export NGINX_VERSION=1.10.1 SIGNING_KEY=A1C052F8 VTS_VERSION=0.1.8 BUILD_PATH=/tmp/build PREFIX=/opt/router \
+	&& export NGINX_VERSION=1.11.2 SIGNING_KEY=A1C052F8 VTS_VERSION=0.1.10 BUILD_PATH=/tmp/build PREFIX=/opt/router \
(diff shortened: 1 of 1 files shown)


deis-admin:

krancour:
@helgi or others. Does anyone have a concern about switching from the stable to mainline branch? I have mixed feelings about it.

helgi:
@krancour no, we always were on mainline, the stable switch only happened when that happened to be the latest

krancour:
I see. That eliminates my only reservation then.

helgi:
We could stay on stable from now on, was only pointing out history :)

bacongobbler:
Jenkins, add to whitelist.

PR #224, opened 2016-08-06
feat(router): Get the annotations from the deployment object instead of RC

kmala:


krancour:
@kmala does this need to be coordinated with a deis/charts change?

kmala:
deis/charts already started using deployments https://github.com/deis/charts/pull/316

krancour:
@kmala that is interesting... I am wondering how it is that everything hasn't broken if the chart uses a deployment for router while router (before this PR) looks for annotations upon its own _rc_.

kmala:
because we still have deprecated RC's still laying around as part of the charts

krancour:
Got it.

sstarcher:
I learned about the requirement of the rc still existing in v.2.3 the hard way.

PR #232, opened 2016-08-11
feat(maintenance): Add support for maintenance mode for apps

kmala:


nginx/config.go
```diff
@@ -188,7 +188,8 @@ http {
 		vhost_traffic_status_filter_by_set_key {{ $appConfig.Name }} application::*;
 
 		location / {
-			{{ if $appConfig.Available }}proxy_buffering off;
+			{{ if $appConfig.Maintenance }}return 503 {{ $appConfig.MaintenancePage }};{{ else if $appConfig.Available }}
+			proxy_buffering off;
```
krancour: Nit... if we care about how the generated configuration looks, we might want to move `proxy_buffering off` up to the end of the previous line. That will avoid an empty line after `location / {` in the event that `$appConfig.Available` is `true`.

model/model.go
```diff
@@ -103,15 +103,17 @@ func newGzipConfig() *GzipConfig {
...
+	Domains         []string `key:"domains" constraint:"(?i)^((([a-z0-9]+(-*[a-z0-9]+)*)|((\\*\\.)?[a-z0-9]+(-*[a-z0-9]+)*\\.)+[a-z0-9]+(-*[a-z0-9]+)+)(\\s*,\\s*)?)+$"`
+	Whitelist       []string `key:"whitelist" constraint:"^((([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])\\.){3}([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])(\\/([0-9]|[1-2][0-9]|3[0-2]))?(\\s*,\\s*)?)+$"`
+	ConnectTimeout  string   `key:"connectTimeout" constraint:"^[1-9]\\d*(ms|[smhdwMy])?$"`
+	TCPTimeout      string   `key:"tcpTimeout" constraint:"^[1-9]\\d*(ms|[smhdwMy])?$"`
+	ServiceIP       string
+	CertMappings    map[string]string `key:"certificates" constraint:"(?i)^((([a-z0-9]+(-*[a-z0-9]+)*)|((\\*\\.)?[a-z0-9]+(-*[a-z0-9]+)*\\.)+[a-z0-9]+(-*[a-z0-9]+)+):([a-z0-9]+(-*[a-z0-9]+)*)(\\s*,\\s*)?)+$"`
+	Certificates    map[string]*Certificate
+	Available       bool
+	Maintenance     bool   `key:"maintenance" constraint:"(?i)^(true|false)$"`
+	MaintenancePage string `key:"maintenancePage"`
```
bacongobbler: is `MaintenancePage` something that has to be embedded into the router? If so I don't think this is a flag we should support, simply because the user would have to make their own changes in the router anyways.

kmala: `MaintenancePage` can be any external Url.

krancour: @bacongobbler I think it's any URL-- could be for someplace completely off cluster-- e.g. a separate subdomain like `maintenance.my-app.com`.

helgi: The first iteration was supposed to be a deis provided `Maintenance Mode` page and a second iteration was supposed to be about bringing in application specific one

bacongobbler: cool, perhaps this should be documented in the readme?

kmala: Yes i can default the value to be deis provided page if its not provided in the annotations but what would deis provided page be will it be an external url or some page in the local file system

helgi: local file - We haven't really hashed out the 2nd iteration (or phase) so it would be best if you just assumed a static local file for now and left all other work for later.

The idea (my rough idea) is it would be a `path` on the app and the router would cache that page locally at X interval or the Controller would ping the router to cache it on every release, or something... with TTLs, etc - Didn't see it as "download any full url" to prevent downloads of big resources or malicious behaviour

README.md
```diff
@@ -287,6 +287,7 @@ _Note that Kubernetes annotation maps are all of Go type `map[string]string`.  A
 | <a name="app-whitelist"></a>routable application | service | [router.deis.io/whitelist](#app-whitelist) | N/A | Comma-delimited list of addresses permitted to access the application (using IP or CIDR notation).  These may either extend or override the router-wide default whitelist (if defined).  Requests from all other addresses are denied. |
 | <a name="app-connect-timeout"></a>routable application | service | [router.deis.io/connectTimeout](#app-connect-timeout) | `"30s"` | nginx `proxy_connect_timeout` setting expressed in units `ms`, `s`, `m`, `h`, `d`, `w`, `M`, or `y`. |
 | <a name="app-tcp-timeout"></a>routable application | service | [router.deis.io/tcpTimeout](#app-tcp-timeout) | router's `defaultTimeout` | nginx `proxy_send_timeout` and `proxy_read_timeout` settings expressed in units `ms`, `s`, `m`, `h`, `d`, `w`, `M`, or `y`. |
+| <a name="app-maintenance"></a>routable application | service | [router.deis.io/maintenance](#app-maintenance) | `False` | Whether the app is under maintenance so that all traffic for this app is redirected to a maintenance page if provided. |
```
krancour: For consistency with other rows in the table, false should be lower case and enclosed in quotes (as well as monospaced as you currently have it. i.e. `"false"`

The quotes are tremendously important, because without them, k8s actually tries to put a `bool` into the annotation map and the annotation map is a `map[string]string`.

krancour: Also:

> Whether the app is under maintenance so that all traffic for this app is redirected to a maintenance page if provided.

After your recent amendments to this PR, that is no longer true... this will _just_ return a 503, without using any maintenance page.

kmala: Updated.

PR #234, opened 2016-08-12
fix(model): clear AppConfigs on build

### LINKED ISSUE #233: attaching `router.deis.io/routable: "false"` still makes the app routable
Given the following service and pod:

```
><> k --namespace=go get svc go -o yaml | grep routable
    router.deis.io/routable: "false"
><> k --namespace=go get po
NAME                      READY     STATUS    RESTARTS   AGE
go-web-1352584931-xvvl9   1/1       Running   0          3m
```

If I attempt to hit the router for that app, it still accepts traffic:

```
><> cat /etc/hosts | grep go
10.245.1.3	k8s.local deis.k8s.local deis-builder.k8s.local deis-workflow-manager.k8s.local deis-minio.k8s.local go.k8s.local foo.k8s.local bar.k8s.local car.k8s.local star.k8s.local www.k8s.local
><> curl go.k8s.local
Powered by Deis
Release v5 on go-web-1352584931-xvvl9
```

Is there a value I can pass to the annotation to disable routing, or do I have to completely remove the annotation from the service to drop requests?

bacongobbler:
When appending to an existing list of AppConfigs, we cannot remove apps from the routing list. Clearing it out on `build()` refreshes the list.

closes #233


model/model.go
```diff
@@ -287,7 +287,7 @@ func build(kubeClient *client.Client, routerDeployment *extensions.Deployment, p
 			return nil, err
 		}
 		if appConfig != nil {
-			routerConfig.AppConfigs = append(routerConfig.AppConfigs, appConfig)
+			routerConfig.AppConfigs = appConfig
```
kmala: types mismatch

bacongobbler: Still waking this morning and forgot to test locally before pushing this out. Thanks

krancour:
Wait... I don't think this actually solves #233

bacongobbler:
@krancour I just amended this. Do you think this will or is there something else in play here?

krancour:
I think there's something else... I believe the root cause is we're actually selecting services we don't intend to when we query the apiserver.

krancour:
I am trying to confirm that with additional debug output.

bacongobbler:
closing in favour of #235.

PR #237, opened 2016-08-19
feat(router): add app SSL config

### LINKED ISSUE #148: Enforce SSL on app-by-app basis
Currently, a cluster admin (if you subscribe to the Google operational model) can enable / enforce redirects from HTTP to HTTPS _router-wide_, but it seems that it would be tremendously useful to app admins (again, if you subscribe to the Google operational model) to opt into that on an app-by-app basis _if_ their cluster admin isn't already enforcing it router-wide.

bacongobbler:
Allow an application to enforce SSL by annotating `router.deis.io/nginx.ssl.enforce` on their service.

closes #148

TODO

 - [x] manual test
 - [x] README updates

### DIFF
--- a/nginx/config.go
+++ b/nginx/config.go
@@ -201,7 +201,7 @@ http {
 			proxy_set_header Upgrade $http_upgrade;
 			proxy_set_header Connection $connection_upgrade;
 
-			{{ if $enforceHTTPS }}if ($access_scheme != "https") {
+			{{ if or $enforceHTTPS $appConfig.SSLConfig.Enforce }}if ($access_scheme != "https") {
 				return 301 https://$host$request_uri;
 			}{{ end }}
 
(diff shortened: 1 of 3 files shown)


nginx/config.go
```diff
@@ -201,7 +201,7 @@ http {
 			proxy_set_header Upgrade $http_upgrade;
 			proxy_set_header Connection $connection_upgrade;
 
-			{{ if $enforceHTTPS }}if ($access_scheme != "https") {
+			{{ if or $enforceHTTPS $appConfig.SSL.Enforce }}if ($access_scheme != "https") {
 				return 301 https://$host$request_uri;
```
felixbuenemann: Shouldn't the redirect keep the query string? That can be done like shown below.

```
return 301 https://$host$request_uri$is_args$args;
```

nginx/config.go
```diff
@@ -201,7 +201,7 @@ http {
 			proxy_set_header Upgrade $http_upgrade;
 			proxy_set_header Connection $connection_upgrade;
 
-			{{ if $enforceHTTPS }}if ($access_scheme != "https") {
+			{{ if or $enforceHTTPS $appConfig.SSL.Enforce }}if ($access_scheme != "https") {
```
felixbuenemann: Doesn't this mean that redirect can't be disabled on a per app basis, if it is enabled globally?

bacongobbler: That is correct and is intentional by design. The requirements are that if the router enforces something, it is not override-able by the user.

Precedence for this behaviour exists at https://github.com/deis/router/blob/master/nginx./config.go#L182

nginx/config.go
bacongobbler: I'll leave that for @krancour to answer as I'm not as familiar with the router.

felixbuenemann: Yeah, it's not directly related to this PR, I just noticed it when reading the diff.

bacongobbler:
Manual test:

```
><> curl go.k8s.local
Powered by Deis
Release v2 on go-web-4166859712-o5f90

><> k --namespace=go edit svc
service "go" edited

><> k --namespace=go get svc -o yaml | grep enforce
      router.deis.io/ssl.enforce: "True"

><> curl go.k8s.local
<html>
<head><title>301 Moved Permanently</title></head>
<body bgcolor="white">
<center><h1>301 Moved Permanently</h1></center>
<hr><center>nginx/1.11.2</center>
</body>
</html>
```

PR #239, opened 2016-08-19
[WIP] feat(model): allow services to bind to multiple ports

bacongobbler:
This feature allows an application to bind to multiple ports on the router. The router now reads
all available ports for a service and binds those ports to the virtual host as necessary. No
changes are required for deis/controller or the application services, since they only expose port
80 in the service, which binds to the container port.

BREAKING CHANGES:

 - the router now runs as root. This is to allow applications to bind to ports lower than 1024, and
   allows the router to not have to maintain a port mapping between port 80 in the service and port
   8080 in the router. This *might* be a breaking change for the ELB on an upgrade because changes
   to the router service will occur.

To address concerns about running nginx as root, users have been running nginx as root in
production for the past decade, ever since nginx workers would run as the "nobody" user, originally
introduced in v0.1.0 in 2004. In a secure setup, a hacker would not be able to access the parent
process and utilize the root account. This is not always true as a hacker could utilize some kind
of exploit to gain root access, but these CVEs are very far and few between, nginx has always
disclosed these security issues and released patches as necessary and as soon as possible. Even
the official nginx image which is maintained by nginx developers runs as root, which makes me
believe that the attack surface is very low.

references deis/workflow#324, and code inspiration originally came from #203.


bacongobbler:
Unfortunately this feature is blocked until https://github.com/kubernetes/kubernetes/issues/23864 is implemented.

nginx/config.go
```diff
@@ -158,15 +150,15 @@ http {
...
-	{{range $appConfig := $routerConfig.AppConfigs}}{{range $domain := $appConfig.Domains}}server {
-		listen 8080{{ if $routerConfig.UseProxyProtocol }} proxy_protocol{{ end }};
+	{{range $appConfig := $routerConfig.AppConfigs}}{{range $domain := $appConfig.Domains}}{{ range $port := $appConfig.ServicePorts }}server {
+		listen {{ $port }}{{ if $routerConfig.UseProxyProtocol }} proxy_protocol{{ end }};
 		server_name {{ if contains "." $domain }}{{ $domain }}{{ else if ne $routerConfig.PlatformDomain "" }}{{ $domain }}.{{ $routerConfig.PlatformDomain }}{{ else }}~^{{ $domain }}\.(?<domain>.+)${{ end }};
 		server_name_in_redirect off;
 		port_in_redirect off;
 		set $app_name "{{ $appConfig.Name }}";
 
 		{{ if index $appConfig.Certificates $domain }}
```
bacongobbler: Note to self: need to handle two cases here:

1. that the port is port 80 (the web port)
2. that the port is not port 443 (since we're already asking the app to listen to port 443 and assume the user is handling certs themselves from within the app)

krancour:
I do still have some concerns about running as root. I agree that running the Nginx master process as root is common and generally safe. The concern I have is oriented more around Docker than it is Nginx. Let's discuss on Monday?

krancour:
@bacongobbler, I'll state my objections, but I also have a solution that probably makes both of us happy.

My objection to running containerized processes as root is that it means users who are able to exec into such a container assume root privileges when they do so. This gives them unchecked access to the container's kernel-- which is shared with the host and all other containers on that host.

This is a big deal in industries where strict separation of responsibilities is dictated by regulatory requirements. Consider, for instance, that within the Google operational model, it is probably only "infra ops" that should have access to hosts, while "cluster ops" can likely use `kubectl` to exec into most any container. Given what I've described above, the existence of containerized processes running as root provides "cluster ops" with a backdoor into "infra ops'" domain.

That being said... some _other_ Workflow components do currently run as root. And further, Workflow offers no guarantee that applications deployed _via_ the platform don't run as root. (fwiw, I feel we should actually address the former outright, and the latter as well, but put that behind a feature flag.) So... I acknowledge that whether the router runs as root or not is nearly moot given that other containers may.

That being said... (a phrase I seem to overuse), router doesn't run as root currently, so I feel strongly that regardless of what other containers are doing, we shouldn't be moving the router in the opposite direction of that best practice it's already implementing.

But... I have a solution...

If I understand correctly, the only compelling reason to run router as root here is the ability to bind to privileged ports. As it turns out (and I just learned this), Docker has a mechanism whereby that capability can be exposed _without_ requiring the containerized process to run as root. i.e. This should satisfy us both.

Here is Docker's documentation on the matter:

https://docs.docker.com/engine/security/security/#/linux-kernel-capabilities

Applicable excerpt:

> Processes (like web servers) that just need to bind on a port below 1024 do not have to run as root: they can just be granted the net_bind_service capability instead. And there are many other capabilities, for almost all the specific areas where root privileges are usually needed.

Kubernetes' docs on the same matter:

http://kubernetes.io/docs/user-guide/containers/

The documentation could stand to give some examples. If I'm not mistaken, you achieve this by adding something like this to the container spec in the manifest:

```
securityContext:
  capabilities:
  - add:
    - NET_BIND_SERVICE
```

Note that must be applied at the _container_ leve; not the pod level.

One final disclaimer-- I have not tried this, but I have observed this trick in use by other OSS projects.

Let me know what you think.

bacongobbler:
> My objection to running containerized processes as root is that it means users who are able to exec into such a container assume root privileges when they do so. This gives them unchecked access to the container's kernel-- which is shared with the host and all other containers on that host.

> That being said... (a phrase I seem to overuse), router doesn't run as root currently, so I feel strongly that regardless of what other containers are doing, we shouldn't be moving the router in the opposite direction of that best practice it's already implementing.

So if I understand our arguments correctly, you're concerned about CVEs/root level access through nginx exploits, and rightfully so. I'm concerned about the ability to bind to ports 1024 and below, which nginx cannot do without running as root.

To address that, the suggestion is to give the nginx container NET_BIND_SERVICE capability instead. We will need to `setcap` the nginx binary as well, as documented [here](https://github.com/docker/docker/issues/8460#issuecomment-87123934). The problem with that approach is that it will not work on AUFS at all as noted [here](https://github.com/docker/docker/issues/8460#issuecomment-126022614). Because we assume we support all k8s setups, this means we can no longer support k8s clusters with AUFS as the backing storage driver for docker. I assume most have moved to OverlayFS at this point, but it is still a concern. If this isn't an issue then I'd be happy to move forward with this approach.

Personally I'm not too concerned about running nginx as root. While the master process runs as root, the worker processes run by default as [the `nobody` user](http://nginx.org/en/docs/ngx_core_module.html#user), meaning incoming requests have to exploit the main nginx process from the worker thread to gain root access. Even [nginx developers have no problem with this approach](https://github.com/nginxinc/docker-nginx/issues/27#issuecomment-111104386). As for users exec'ing into the container... They are assumed to be administrators of the cluster as they already have `kubectl` access. Furthermore Nginx Inc. has shown time and again that they reasonably disclose CVE exploits with high severity (such as remote exec capabilities), demonstrating which releases are affected and release patches/new releases as necessary. Even if an exploit is found the turnaround for us to update nginx is usually within an hour.

I'll experiment with the NET_BIND_SERVICE capability on my own free time. This feature is still blocked until port ranges can be used for the router service manifest via https://github.com/kubernetes/kubernetes/issues/23864 so there's no rush on this feature.

krancour:
> So if I understand our arguments correctly, you're concerned about CVEs/root level access through nginx exploits

Nope. Altho that may be a valid concern, I am more concerned about users who can `kubectl exec`.

> I'm concerned about the ability to bind to ports 1024 and below

A valid concern, for sure. I hope I'm not trivializing that. I completely understand your motivations.

> As for users exec'ing into the container... They are assumed to be administrators of the cluster as they already have kubectl access.

That statement doesn't necessarily hold up at a large company, especially in a regulated industry. People who have direct access to hosts and those who have fairly broad access to all containers via `kubectl`.

Also, we're accustomed to working with k8s clusters where RBAC isn't really configured or enforced. For anyone who's put care and effort into that, as they should, there may be a class of users with limited permissions who, for instance, are able to exec into containers within their own app's namespace only. Such users definitely _aren't_ meant to have any access to the host's kernel.

To address technical points:

> We will need to setcap the nginx binary as well, as documented here.

I assumed (perhaps incorrectly) that Docker's `--add-cap` flag (which is what k8s is using under the hood) accomplished this without any setcap being required. (fwiw, the router did use setcap once upon a time and I pulled it out because it added to much heft to the image size.) I think it's worth experimenting with, and I can spend 10 or 15 minutes as well to see if we need to explicitly setcap or not.

The AUFS-related limitation is news to me. Good find there. I am wondering if it still constrains us if, as I am hoping, Docker's `--add-cap` can fully replace a `setcap` within the container, as I hope it can. Again-- perhaps only experimentation will tell.

I at least feel like we're converging on a common understanding of each other's point of view here and a _possible_ compromise that might be just a smidge out of reach. Let's find out for sure.

PR #248, opened 2016-09-09
fix(ngnix): handle X-Forwarded-Port and X-Forwarded-Proto properly

### LINKED ISSUE #244: Deis Router does not allow for X-Forwarded-Proto to be "wss"
The deis router does not allow for a currently existing X-Forwarded-Proto to be "wss" for websockets. It instead rewrites the proto to be "http"

kmala:
fixes #244

### DIFF
--- a/nginx/config.go
+++ b/nginx/config.go
@@ -75,12 +75,15 @@ http {
 		default $scheme;               # if X-Forwarded-Proto header is empty, $tmp_access_scheme will be the actual protocol used
 		"~^(.*, ?)?http$" "http";      # account for the possibility of a comma-delimited X-Forwarded-Proto header value
 		"~^(.*, ?)?https$" "https";    # account for the possibility of a comma-delimited X-Forwarded-Proto header value
+		"~^(.*, ?)?ws$" "ws";      # account for the possibility of a comma-delimited X-Forwarded-Proto header value
+		"~^(.*, ?)?wss$" "wss";    # account for the possibility of a comma-delimited X-Forwarded-Proto header value
 	}
-	# 2. If the request is an HTTPS request, upgrade $access_scheme to https, regardless of what the X-Forwarded-Proto
+	# 2. If the request is an HTTPS/wss request, upgrade $access_scheme to https/wss, regardless of what the X-Forwarded-Proto
 	# header might say.
 	map $scheme $access_scheme {
 		default $tmp_access_scheme;
 		"https" "https";
+		"wss"	"wss";
 	}
 
 	# Determine the forwarded port:
@@ -94,10 +97,16 @@ http {
 	# 2. If the X-Forwarded-Port header has been set already (e.g. by a load balancer), use its
 	# value, otherwise, the port we're forwarding for is the $standard_server_port we determined
 	# above.
-	map $http_x_forwarded_proto $forwarded_port {
+	map $http_x_forwarded_port $forwarded_port {
 		default $http_x_forwarded_port;
 		'' $standard_server_port;
 	}
+	# uri_scheme will be the scheme to use when the ssl is enforced.
+	map $access_scheme $uri_scheme {
+		default "https";
+		"ws"	"wss";
+	}
+
 
 	{{ $sslConfig := $routerConfig.SSLConfig }}
 	{{ $hstsConfig := $sslConfig.HSTSConfig }}{{ if $hstsConfig.Enabled }}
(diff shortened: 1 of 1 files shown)


krancour:
I can't review this atm, but I'd like to look at this one in some more detail before we merge, if that's alright.

kmala:
@krancour this isn't getting merged without your lgtm

nginx/config.go
```diff
@@ -73,8 +73,9 @@ http {
 	# treated as if it were HTTPs.
 	map $http_x_forwarded_proto $tmp_access_scheme {
 		default $scheme;               # if X-Forwarded-Proto header is empty, $tmp_access_scheme will be the actual protocol used
-		"~^(.*, ?)?http$" "http";      # account for the possibility of a comma-delimited X-Forwarded-Proto header value
-		"~^(.*, ?)?https$" "https";    # account for the possibility of a comma-delimited X-Forwarded-Proto header value
+		"~^(.*, ?)?https(,.* ?)?$" "https";      # account for the possibility of a comma-delimited X-Forwarded-Proto header value to see if https is present
+		"~^(.*, ?)?http(,.* ?)?$" "http";      # account for the possibility of a comma-delimited X-Forwarded-Proto header value to see if http is present
```
krancour: I had to put a lot of though into this and in the end, I'm not entirely sure about loosening up these two regexes like this.

Determining the access scheme is an exercise in determining when a request should be treated as an HTTPS request even though SSL offloading might already have occurred.

Before, the access scheme was considered HTTPS if (and only if) the potentially coma-delimited list of forwarded protocols _ended_ with HTTPS. That would be indicative of SSL having been offloaded by the most proximate proxy (usually your load balancer).

This change treats a request as HTTPS if _any_ intermediate proxy offloaded SSL (or claims to have done so).

Here's the critical question: If SSL were offloaded by some other proxy _before_ it reached your load balancer, should we really regard that request as secure? Since this means an unencrypted request may have traversed untrusted networks, I would argue no... and that's _without_ any malicious actors in the mix. If you assume there could be, there's no shortage of creative means by which I can think of abusing this.

Let me ask... how critical is loosening these regexes to fixing #244? Unless I am mistaken, we could do without these two changes. Thoughts?

nginx/config.go
```diff
@@ -73,8 +73,9 @@ http {
...
+		"~^[A-Za-z][A-Za-z0-9+.-]*" $http_x_forwarded_proto; # Use the value passed in the header if its a valid URI scheme
```
krancour: > Use the value passed in the header if its a valid URI scheme

Won't this use the passed in `$http_x_forwarded_proto` even if it's _not_ a valid scheme?

nginx/config.go
kmala: i can revert these.

nginx/config.go
kmala: no..if its not a valid scheme then `default  $scheme` will be passed.

krancour: Where is the logic that's determining if the scheme is valid or not?

kmala: the regex is based on the rfc for a valid scheme name https://www.ietf.org/rfc/rfc2396.txt(section 3.1) and yes it accepts all the scheme names both which are accepted by IANA http://www.iana.org/assignments/uri-schemes/uri-schemes.xhtml and the private schemes.
so here my question is do we want to be loose on what needs to be accepted or more strict and add/handle them upon user requests(currently the request is only for web sockets)?
i had actually forgot about the EnforceSSL part, websockets have two schemes ws:// and wss:// similar to http and if enforcessl is present we need to change the scheme from ws to wss.

krancour: @kmala ok. Didn't know the regex was based off an RFC. If that regex describes what the RFC considers valid, cool, but I guess my question is more about validity with respect to what touter can actually handle. What happens, for instance, when the scheme is `xyz`?

Re: `ws` vs `wss`-- thank you for explaining that-- I didn't even know wss was a thing. I really _do_ need to research web sockets more. Do we need to specially account for these schemes in the way we do http and https so we can correctly determine the scenarios where we need to enforce SSL by redirecting?

kmala: yes...that can still be a private scheme internally used by someone, hence my question do we want to be more strict or loose on these things?I assume more strict and just add support for web sockets for now as those are the only things apart from http which most people will use.

krancour: I think that sounds reasonable.

bacongobbler:
pinging @krancour for a re-review before we cut a new router release

kmala:
@krancour i had tested with an example app based on https://www.nginx.com/blog/websocket-nginx/ to test the web sockets

PR #252, opened 2016-09-13
chore(nginx): update nginx to 1.11.4

helgi:
https://nginx.org/en/CHANGES


mboersma:
Nice! I tried updating nginx to 1.11.3 but couldn't get streams to compile. This explains why.

helgi:
That's what we get for being on the bleeding edge :-) We may want to revisit that practice in the future, tho I do believe I may have been the one that moved us from 1.10 to 1.11 ;-)

PR #256, opened 2016-09-22
fix(Dockerfile): update openssl to 1.0.2i

bacongobbler:
See https://www.openssl.org/news/secadv/20160922.txt


helgi:
Compiling OpenSSL out of band from Ubuntu never turns out well. If we need to get things updated  (see http://packages.ubuntu.com/xenial/openssl - the patch isn't there yet, http://changelogs.ubuntu.com/changelogs/pool/main/o/openssl/openssl_1.0.2g-1ubuntu4.1/changelog - `g` part doesn't matter since they patch in place)

The high sev there is around OCSP stapling and we don't have that enabled (even if we should, created a ticket around that): https://github.com/deis/router/search?utf8=%E2%9C%93&q=ssl_stapling&type=Code - The rest of the tickets were sometimes not relevant (DTLS for example) and then other little smatterings of old ciphers and such.

We should update the base image instead when the release is out or specifically pull in openssl updates on router, thoughts?

helgi:
Lets let this through, @bacongobbler will give some color as to why - Also lets create ticket that removes this compile stuffer when Ubuntu has released a newer OpenSSL

bacongobbler:
We have a high-priority ticket to upgrade both the router and heroku/cedar:14 with 1.0.2i and 1.0.1u. As of now heroku/cedar:14 is on 1.0.1f so we'll have to submit a PR upstream to update that image.

krancour:
OpenSSL seems to be included in our quay.io/deis/base image. Shouldn't we fix this there and _any_ of our images (router included) that would benefit from that update can just step up to the latest base?

bacongobbler:
I don't see an explicit callout to openssl at https://github.com/deis/docker-base/blob/master/rootfs/Dockerfile#L9-L12. I'd be happy to implement it there if you'd prefer.

What's odd is that Ubuntu has not yet provided an update on apt for openssl so most users can't `apt-get update && apt-get upgrade`.

rootfs/Dockerfile
```diff
@@ -17,11 +17,18 @@ RUN apt-get update \
...
 		make \
-	&& export NGINX_VERSION=1.11.4 SIGNING_KEY=A1C052F8 VTS_VERSION=0.1.10 BUILD_PATH=/tmp/build PREFIX=/opt/router \
+	&& export NGINX_VERSION=1.11.4 OPENSSL_VERSION=1_0_2i SIGNING_KEY=A1C052F8 VTS_VERSION=0.1.10 BUILD_PATH=/tmp/build PREFIX=/opt/router \
 	&& rm -rf "$PREFIX" \
 	&& mkdir "$PREFIX" \
 	&& mkdir "$BUILD_PATH" \
 	&& cd "$BUILD_PATH" \
+	&& get_src 7e07311e7a05750e015f78b216c6ec9e4596a21a50eb93d3dc7a608b9900b3c0 \
+		"https://github.com/openssl/openssl/archive/OpenSSL_$OPENSSL_VERSION.tar.gz" \
+	&& cd "$BUILD_PATH/openssl-OpenSSL_$OPENSSL_VERSION" \
```
mboersma: You could use `pushd` instead of `cd` here, and `popd` instead of `cd ..` down below. Then this code doesn't assume as much about the directory nesting level.

bacongobbler: good idea. I'll probably implement the same thing down lower.

bacongobbler:
Right now I'm holding off on merging this because I want to wait for a release on apt for openssl so we don't require maintaining this.

bacongobbler:
From reading up on https://www.debian.org/security/2016/dsa-3673:

> For the unstable distribution (sid), these problems will be fixed soon.

```
$ docker run quay.io/deisci/router:canary cat /etc/debian_version
stretch/sid
```

So we have to wait patiently for an update or downgrade our docker-base image to debian jessie.

bacongobbler:
closing in favour of upgrading deis/docker-base in deis/docker-base#11

PR #259, opened 2016-09-23
ref(glide): use the official kubernetes client

arschles:
cc/ @krancour


arschles:
the `vendor/` directory after a fresh `glide install` is now 39.1mb

arschles:
would like to get some manual testing on this, if anyone knows some paths that weren't hit by e2e. otherwise, will merge on monday

PR #260, opened 2016-09-23
chore(Dockerfile): bump deis/base to v0.3.2

bacongobbler:
See https://github.com/deis/docker-base/releases/tag/v0.3.2

### DIFF
--- a/rootfs/Dockerfile
+++ b/rootfs/Dockerfile
@@ -1,4 +1,4 @@
-FROM quay.io/deis/base:0.3.1
+FROM quay.io/deis/base:v0.3.2
 
 RUN adduser --system \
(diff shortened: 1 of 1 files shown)


rootfs/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM quay.io/deis/base:0.3.1
```
jeroenvisser101: All other tags (for the base image) don't include the 'v', is this now adopted platform-wide?

bacongobbler: Yes,

jeroenvisser101: Okay!

PR #262, opened 2016-09-26
chore(Dockerfile): cleanup after installation a bit more than before and keep copyright / license files

helgi:
Intentionally left `--no-install-recommends` out since it was not used already, this will make the image bigger but may also be needed. Can try to add that later if people want me to


helgi:
I introduced `--no-install-recommends` on the PR and all lights are still green by the looks of it!

PR #263, opened 2016-09-27
fix(nginx): Enable builder PROXY PROTOCOL support

### LINKED ISSUE #255: AWS elb Proxy protocol breaking deis/builder
Trying to get the client IP on my app request, I hit an issue with proxy protocol.
I've enable on the AWS elb and on the deis-router, It works fine through nginx,
but If I try to push my app to the deis/builder, it breaks the request.

Since K8s enables proxy protocol to every listener on the service, including the deis/builder.
Not supporting proxy protocol on the deis/builder, might be an issue.

felixbuenemann:
If proxy protocol is enabled on an AWS ELB in kubernetes using the
`service.beta.kubernetes.io/aws-load-balancer-proxy-protocol: '*'`
annotation on the load balancer service, it will enable PROXY PROTOCOL
on all ports forwarded by the ELB, which means the builder proxy on port
2222 also needs to use `proxy_protocol` on its `listen` directive.

There is currently no way to configure PROXY PROTOCOL per port on AWS
using kubenetes service annotation and it doesn't really make sense to
enable proxy protocol only for port 80 and 443 anyways, so this fix uses
the same configuration setting as is used for http and https.

The required `proxy_protocol` option for the `listen` directive was [added in NGINX 1.11.4](https://nginx.org/en/docs/stream/ngx_stream_core_module.html#listen).

This fixes #255.

### DIFF
--- a/nginx/config.go
+++ b/nginx/config.go
@@ -247,7 +247,7 @@ http {
 
 {{ if $routerConfig.BuilderConfig }}{{ $builderConfig := $routerConfig.BuilderConfig }}stream {
 	server {
-		listen 2222;
+		listen 2222 {{ if $routerConfig.UseProxyProtocol }}proxy_protocol{{ end }};
 		proxy_connect_timeout {{ $builderConfig.ConnectTimeout }};
 		proxy_timeout {{ $builderConfig.TCPTimeout }};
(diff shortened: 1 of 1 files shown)


deis-admin:

mboersma:
Jenkins, add to whitelist.

PR #268, opened 2016-10-04
docs(README): Remove `debug` from description in error-log-level.

monaka:
I got the error like this. "debug" is not valid.

```
2016-10-04 00:13:32.353662 I | WARNING: Field "router.deis.io/nginx.errorLogLevel" value "debug" does not satisfy constraint /^(info|notice|warn|error|crit|alert|emerg)$/ -- skipping this field and using default value "error".
```


deis-admin:

helgi:
See https://github.com/deis/router/pull/267 - it adds `debug`

monaka:
I see. It's good to support `debug`. I close this PR.

PR #269, opened 2016-10-05
fix(docs): correct annotation key for maxWorkerConnections

ultimateboy:
See: https://github.com/deis/router/blob/master/model/model.go#L40


krancour:
Nice catch.

krancour:
Sorry for accidentally closing-- hit the wrong button!!!

PR #270, opened 2016-10-05
feat(model): allow configuration through a configmap.

ultimateboy:
While this does function as desired, this PR is still a WIP - mainly needing updated docs.

### DIFF
--- a/model/model.go
+++ b/model/model.go
@@ -24,9 +24,10 @@ const (
 )
 
 var (
-	namespace   = utils.GetOpt("POD_NAMESPACE", "default")
-	modeler     = modelerUtility.NewModeler(prefix, modelerFieldTag, modelerConstraintTag, true)
-	listOptions api.ListOptions
+	namespace         = utils.GetOpt("POD_NAMESPACE", "default")
+	annotationModeler = modelerUtility.NewModeler(prefix, modelerFieldTag, modelerConstraintTag, true)
+	configmapModeler  = modelerUtility.NewModeler("", modelerFieldTag, modelerConstraintTag, true)
+	listOptions       api.ListOptions
 )
 
 func init() {
@@ -209,14 +210,27 @@ func newHSTSConfig() *HSTSConfig {
 // relevant metadata concerning itself and all routable services.
 func Build(kubeClient *kubernetes.Clientset) (*RouterConfig, error) {
 	// Get all relevant information from k8s:
+	//   deis-router configMap (if exists)
 	//   deis-router deployment
 	//   All services with label "routable=true"
 	//   deis-builder service, if it exists
 	// These are used to construct a model...
+
+	configMap, err := getConfigMap(kubeClient)
+	if err != nil {
+		if !errors.IsNotFound(err) {
+			// If the error is anything but not found, fail.
+			return nil, err
+		}
+		// Otherwise, the configMap is not found, set as empty.
+		configMap = nil
+	}
+
 	routerDeployment, err := getDeployment(kubeClient)
 	if err != nil {
 		return nil, err
 	}
+
 	appServices, err := getAppServices(kubeClient)
 	if err != nil {
 		return nil, err
@@ -235,7 +249,7 @@ func Build(kubeClient *kubernetes.Clientset) (*RouterConfig, error) {
 		return nil, err
 	}
 	// Build the model...
-	routerConfig, err := build(kubeClient, routerDeployment, platformCertSecret, dhParamSecret, appServices, builderService)
+	routerConfig, err := build(kubeClient, routerDeployment, configMap, platformCertSecret, dhParamSecret, appServices, builderService)
 	if err != nil {
 		return nil, err
 	}
@@ -248,6 +262,14 @@ func getDeployment(kubeClient *kubernetes.Clientset) (*v1beta1ext.Deployment, er
 		return nil, err
 	}
 	return deployment, nil
+}
+
+func getConfigMap(kubeClient *kubernetes.Clientset) (*v1.ConfigMap, error) {
+	configMap, err := kubeClient.ConfigMaps(namespace).Get("deis-router")
+	if err != nil {
+		return nil, err
+	}
+	return configMap, nil
 }
 
 func getAppServices(kubeClient *kubernetes.Clientset) (*v1.ServiceList, error) {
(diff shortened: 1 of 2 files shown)


krancour:
I've actually been wanting to do this. I'll take a look.

ultimateboy:
A sample configmap:

```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: deis-router
  namespace: deis
  labels:
    heritage: deis
data:
  nginx.errorLogLevel: "info"
  nginx.maxWorkerConnections: "767"
```

model/model.go
```diff
@@ -317,12 +340,23 @@ func build(kubeClient *kubernetes.Clientset, routerDeployment *v1beta1ext.Deploy
...
+	if configMap.Data != nil {
+		err := configmapModeler.MapToModel(configMap.Data, "nginx", routerConfig)
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	// To maintain backwards compatibility with previous Deis releases and Kube < 1.2
+	// annotations on the router Deployment override the ConfigMap.
+	err := annotationModeler.MapToModel(routerDeployment.Annotations, "nginx", routerConfig)
```
krancour: Two things here:

First, we really need to verify that these two sources of configuration can be merged cleanly. Consider (as one for instance) that a `RouterConfig` instance has a nested `GzipConfig` (or `SSLConfig`). If the second source of configuration (annotations) override _one_ attribute of one of those nested objects, we need to verify that it doesn't blank out all the _other_ attributes of those nested objects. (This is a _classic_ problem in merging maps of maps.) I _think_ it works ok in this particular case, but we really should double check.

The second thing (which might obviate the first)... do we really want to merge the two sources of configuration? I don't like the "config spray" where config values affecting the model could be found in so many different places. It might be simpler from an operator's perspective if this were as simple as "configuration comes from the configMap if it exists, else it comes from annotations; no merging of the two takes place."

krancour:
I've made some line comments that I think should be addressed. Those aside, I want to make a more general comment as well.

I previously said:

> I've actually been wanting to do this.

I'm now second guessing my own impulse. `ConfigMaps` are sexy. They seem like __the right way™__ to configure things, but annotations also make a high degree of sense here and seem pretty defensible as well-- if routable services are configured through annotations, why not configure the router in the same way? It's consistent.

I realize that (mostly for the sake of backward compatibility), this new feature of managing router config through a `ConfigMap` doesn't preclude the the possibility of doing so with annotations. What it does do, however, is increase complexity for an operator who's hunting down an issue by adding multiple sources of configuration. It also would add complexity to the documentation. (Since `ConfigMap` keys disallow the `/` character, as you have noted, the `router.deis.io/` prefix on keys needs to be dropped. These seems to add a complex new dimension to configuration, the explanation for which will be too easily overlooked in the documentation, which will doubtless give some operators difficulties that we will need to assist them with.)

So...

I'm not saying we shouldn't do this. I'm not. But, if we were to do this, I think we need to have a really compelling answer for why. What is the tangible benefit of doing so?

model/model.go
```diff
@@ -209,14 +210,28 @@ func newHSTSConfig() *HSTSConfig {
...
+
+	configMap, err := getConfigMap(kubeClient)
+	if err != nil {
+		statusErr, ok := err.(*errors.StatusError)
+		if ok && statusErr.Status().Code != 404 {
+			// If the error is anything but not found, fail.
+			return nil, err
+		}
+		// Otherwise, the configMap is not found, set as empty.
+		configMap = &v1.ConfigMap{}
```
ultimateboy: Fixed.

model/model.go
```diff
@@ -209,14 +210,28 @@ func newHSTSConfig() *HSTSConfig {
...
+	}
+
```
ultimateboy: Fixed.

model/model.go
```diff
@@ -317,12 +340,23 @@ func build(kubeClient *kubernetes.Clientset, routerDeployment *v1beta1ext.Deploy
 	return routerConfig, nil
 }
 
-func buildRouterConfig(routerDeployment *v1beta1.Deployment, platformCertSecret *v1.Secret, dhParamSecret *v1.Secret) (*RouterConfig, error) {
+func buildRouterConfig(routerDeployment *v1beta1.Deployment, configMap *v1.ConfigMap, platformCertSecret *v1.Secret, dhParamSecret *v1.Secret) (*RouterConfig, error) {
 	routerConfig := newRouterConfig()
-	err := modeler.MapToModel(routerDeployment.Annotations, "nginx", routerConfig)
+
+	if configMap.Data != nil {
```
ultimateboy: Fixed

model/model.go
ultimateboy: I added test coverage to the configuration merge functionality. We can now ensure that the expected RouterConfig is returned when a deployment with annotations is combined with a ConfigMap.

krancour: But as for the question of whether we _should_ merge the two?

Note, this is separate from the question of whether we should support configuration via configmaps _at all_, but both questions have the same motivations-- multiple sources of config (or hierarchical sources of config) make it harder for an operator to comprehend and/or modify the configuration.

ultimateboy: Alright, let's break down the options and pros/cons of each:

1. Configuration done through deployment annotations
  - Pros:
    - No changes, though we'll still need to add test coverage
    - Consistent configuration: you configure apps via annotations, you configure router via annotations
  - Cons:
    - To configure router, operator must edit deployment manifest - most of which is not relevant to the configurations he/she would be making
    - Annotation keys are long with the prefix of "router.deis.io/"

2. Deployment annotations override configurations within a ConfigMap
  - Pros:
    - Using configmaps is the "expected" way to configure an application within Kubernetes
    - Isolates the operator-configurable parts of the router from the rest of the surrounding templated yaml that currently exists in the deployment manifest
  - Cons:
    - Multiple sources of configuration make it harder for an operator to track down where a specific configuration is coming from
    - Because of restrictions on configmap keys, the keys of the configmap and the annotation cannot exactly match, making documentation difficult
      - This is slightly mitigated by focusing the main documentation on the configmap method and link off to a page which describes the annotations override behavior

3. Use the configmap if it exists, otherwise fallback to annotations
  - Pros:
    - Using configmaps is the "expected" way to configure an application within Kubernetes
    - Isolates the operator-configurable parts of the router from the rest of the surrounding templated yaml that currently exists in the deployment manifest
  - Cons:
    - Because of restrictions on configmap keys, the keys of the configmap and the annotation cannot exactly match, making documentation difficult
      - This is slightly mitigated by focusing the main documentation on the configmap method and link off to a page which describes the annotations

4. Only use configmap
  - Pros:
    - Simple
    - Using configmaps is the "expected" way to configure an application within Kubernetes
    - Isolates the operator-configurable parts of the router from the rest of the surrounding templated yaml that currently exists in the deployment manifest
  - Cons:
    - Not compatible with Kube < 1.2
    - Not backwards-compatible with previous versions of Workflow
      - Script to migrate previous annotation-configs to configmap-configs could help mitigate


Based on these, I tend to lean towards number 3, unless of course we could actually get away with number 4 - but I don't feel like we can.

ultimateboy: In the end, I really don't care. I took this on because of the multiple mentions of wanting to transition to using configmaps over annotations. If you've decided you don't like it after all, I'll just close this PR and focus on implementing some test coverage for the annotation configuration method.

jchauncey:
I want to weigh in on this PR but I feel like this needs to be part of a larger architectural discussion. Can we schedule a 30 minute meeting to discuss the details and lay everything out? And try to come to a consensus.

krancour:
@jchauncey absolutely.

PR #271, opened 2016-10-05
feat(charts): Add helm charts for the router

kmala:


helgi:
Do you want to ditch the `manifests` dir now or when we have fully moved over?

kmala:
I will remove it after merging this and ci job to publish the charts along with updating the readme file.

helgi:
Giving it 👍 as long as you are sure this is up to date with the current charts :)

charts/router/Chart.yaml
```diff
@@ -0,0 +1,10 @@
+name: router
+home: https://github.com/deis/router
+version: <Will be populated by the ci before publishing the chart>
+description: Edge router for Deis Workflow.
+maintainers:
+  - name: Deis Team
+    email: engineering@deis.com
+details: |-
```
ultimateboy: These are the details for workflow, not router.

krancour:
@kmala-- this looks pretty good. I have just a couple questions.

1. Is the idea (in general), that each Workflow component will have its own chart sourced embedded within its own repo? Will the Workflow chart reference each of these as dependencies? If that's the case, I think I really like this. 😄 

1. What's the workflow for deploying the router's chart? I imagine each time we tag a release, we want packaging and uploading of the chart to be automatic, right? That doesn't seem to be accounted for here.

1. @helgi mentioned getting rid of the `/manifests` directory soon. That would be great. When that happens, what's the new workflow for a contributor who has (more than likely) installed all of Workflow into a cluster and now wishes to replace _just_ the router component with a custom build? (i.e. You've `helm` installed Workflow and that has resulted in a Workflow release in your cluster, but to hack on router, you want to replace a few discrete _pieces_ of that Workflow release. I'm not sure how that is accomplished.)

kmala:
>1. Is the idea (in general), that each Workflow component will have its own chart sourced embedded within its own repo? Will the Workflow chart reference each of these as dependencies?

yes, each will be in its own repo and workflow will add these add dependencies.
>2.What's the workflow for deploying the router's chart? I imagine each time we tag a release, we want packaging and uploading of the chart to be automatic, right?

 if you want just router you can `helm install charts.deis.com/router-<version>` or you install workflow and the router will be part of it. Tagging a new release will automatically package and upload the chart which will doe through our CI(I am working on it).
>3.What's the new workflow for a contributor who has (more than likely) installed all of Workflow into a cluster and now wishes to replace just the router component with a custom build? 

Currently, you either hack on the manifest through kubectl or install a new workflow with the latest router.But i would want something like `helm upgrade <workflow-release-name> <dependency-name> <chart of the dependency>` to happen and i am discussing with helm team for this use case.

You can see more details on how everything should work here https://github.com/deis/charts/issues/350#issuecomment-251587320

krancour:
> if you want just router you can helm install charts.deis.com/router-<version> or you install workflow and the router will be part of it. Tagging a new release will automatically package and upload the chart which will doe through our CI(I am working on it).

Sorry... I need to restate my question more clearly...

What is the process for taking the chart sources from this repo, archiving them, and uploading them to a repository? Ideally, what is the process for doing this automatically _when a router release gets tagged in github_?

kmala:
When ever we tag a release there is already a job which runs to verify the image and push it to the deis org with the release tag, i want to add to the same job once it uploads the new image replace the org, version in the values.yaml file in the charts directory, run `helm publish <router>` and upload the tar.gz to our hosting directory(s3).

krancour:
Ok. Cool. Should that be part of this PR? Or you want to tackle that separately?

kmala:
that will be part of the our jobs infrastructure... it will be just one job for all the repos.

kmala:
i will merge this PR once that has been setup.

krancour:
Got it.

charts/router/Chart.yaml
kmala: details section is removed in the new helm. So removed it from here.

PR #275, opened 2016-10-06
docs(readme): remove reference to `make dev-registry`

ultimateboy:


krancour:
Ya... that make target disappeared a while ago, which made me kinda sad. I seem to enjoy make more than most. I thought it was a nice convenience having a target that made a local registry for you (if that suited your preferred development workflow), but I guess others saw it as bloat. Regardless... this doc change is good since it eliminates docs for that target, which was most assuredly defunct awhile ago.

PR #279, opened 2016-10-07
fix(nginx): known model shouldn't be refreshed if Nginx reload failed.

### LINKED ISSUE #278: Bug: "known" model shouldn't be refreshed if Nginx reload failed
See following lines:

https://github.com/deis/router/blob/b33c18c768a6adebc0a41a5444852391d2bf4407/router.go#L53-L54

We should be capturing the error here and setting `known = routerConfig` _only_ if reload succeeded (i.e. `err != nil`).

Because we're _not_ doing this, what can happen is that Nginx continues serving requests using old configuration (unless reload triggered a _fatal_ error). While Nginx continues serving requests using old config, Router believes Nginx is up to date with config based on the model the router now considers current. This disagreement over current state would continue until some new change results in an updated model.

cc @monaka @ultimateboy

ultimateboy:
Fixes #278

### DIFF
--- a/router.go
+++ b/router.go
@@ -50,7 +50,11 @@ func main() {
 			log.Printf("Failed to write new nginx configuration; continuing with existing configuration: %v", err)
 			continue
 		}
-		nginx.Reload()
+		err = nginx.Reload()
+		if err != nil {
+			log.Printf("Failed to reload nginx; continuing with existing configuration: %v", err)
+			continue
+		}
 		known = routerConfig
 	}
 }


krancour:
Perfect.

router.go
```diff
@@ -50,7 +50,11 @@ func main() {
 			log.Printf("Failed to write new nginx configuration; continuing with existing configuration: %v", err)
 			continue
 		}
-		nginx.Reload()
+		err = nginx.Reload()
```
ultimateboy: Fair enough. I just followed what the rest of the file was doing. We'll switch over to the one-liner in another PR.

Issue #212, opened 2016-07-02
With many (custom) domains the router might be too slow to startup


nathansamson:
Scenario:
Have a few applications; but many custom domains. (100s of them, I have around 300, the specific amount that triggers the problem will be dependant on the load and CPU/memory size of the node)

Restart the router replica (eg during an upgrade). One way to achieve this is to run
`kubectl --namespace deis scale rc --replicas=0; kubectl --namespace deis scale rc --replicas=1`

Follow the pods status; see it never coming up; after 20s it will go into Restarts 1
After some more time it restarts again and so on.

It seems the initial configuration is taking so long to build that nginx never comes up, and the K8s system decides to restart it. 

manually removing some domans from the annotations of an application made the routers start again

I worked around this issue by changing the router-rc manifest

~~~
        livenessProbe:
          httpGet:
            path: /healthz
            port: 9090
          initialDelaySeconds: 60
          timeoutSeconds: 1
        readinessProbe:
          httpGet:
            path: /healthz
            port: 9090
          initialDelaySeconds: 60
          timeoutSeconds: 1
~~~

Note I can not figure out why it initially restarted after 20s, because the initialDelaySeconds was also 1.

Note that I am not sure what the correct solution would be; a few solutions spring to mind

* Make it faster to generate the config. Obviously this only increases the problem treshold, so unless it is like 10 times faster (which is unlikely this is possible) this would not help much
* Make the default config a working config; obviously this has the disadvantage the router is in an up state while in fact it isn't....
* Increase the healthcheck timeout, but again this only changes the treshold a bit...

krancour:
I think this is to be expected... that if you saddle the router with a masive amount of configuration, the time to build the model will be larger than the timeout on the liveness checks. If you have hundreds of domains, I think tweaking the health checks at the time of installation is a reasonable thing to do. That said, I'll take a look Tuesday to see if there are any other optimizations that can be made to the checks to make them more tolerant of this scenario.

nathansamson:
Note that one of the problems I had investigating the issue was that the logs of the router looked something like this:

~~~
nginx starting...
nginx started
~~~

immediatelly after startup, then after a while the pods stopped working without the pod displaying any logs.
This confused me because it looked like nginx was actually running, and nothing was crashing in the pod, so there was no reason for it to be restarted....


I would suggest to change the logging to something like


~~~
initial nginx starting...
initial nginx started
bulding first config file.... (this can take a while, increase liveness timeout if the application restarts before this finishes)
[...]

PR #276, opened 2016-10-06
tests(model): add test coverage for model

ultimateboy:

PR #280, opened 2016-10-07
fix(model): check if CertMapping exists

bacongobbler:
This way we're not performing a needless API call to kubernetes.

addresses #212 slightly


ultimateboy:
Test coverage for this package is added in https://github.com/deis/router/pull/276 (and therefore I'd ignore the codecov rejection)

PR #281, opened 2016-10-07
tests(nginx): add tests for nginx package.

ultimateboy:


ultimateboy:
Before:
```
?   	github.com/deis/router/nginx	[no test files]
```

After:
```
ok  	github.com/deis/router/nginx	0.034s	coverage: 56.5% of statements
```

PR #288, opened 2016-11-09
feat(charts): Add support to add annotations during install

kmala:
fixes https://github.com/deis/charts/issues/374

This is useful if any user wants to have default annotations for all his cluster installs by using the same values file for all installs thereby not needing to do custom changes after the install.

### DIFF
--- a/charts/router/values.yaml
+++ b/charts/router/values.yaml
@@ -5,3 +5,13 @@ platform_domain: ""
 dhparam: ""
 # limits_cpu: "100m"
 # limits_memory: "50Mi"
+
+# Any custom router annotations(https://github.com/deis/router#annotations)
+# which need to be applied can be specified as key-value pairs under "deployment_annotations"
+deployment_annotations:
+  #<example-key>: <example-value>
+
+# Any custom annotations for k8s services like http://kubernetes.io/docs/user-guide/services/#ssl-support-on-aws
+# which need to be applied can be specified as key-value pairs under "service_annotations"
+service_annotations:
+  #<example-key>: <example-value>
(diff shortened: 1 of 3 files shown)


charts/router/values.yaml
```diff
@@ -5,3 +5,13 @@ platform_domain: ""
 dhparam: ""
 # limits_cpu: "100m"
 # limits_memory: "50Mi"
+
+# Any custom router annotations(https://github.com/deis/router#annotations)
+# which needs to be applied can be specified as key-value pairs under deployment_annotations key
```
mboersma: Minor tweaks:
"needs" -> "need"

Either say `the "deployment_annotations" key` or just `"deployment_annotations"`

charts/router/values.yaml
```diff
@@ -5,3 +5,13 @@ platform_domain: ""
...
 # limits_cpu: "100m"
 # limits_memory: "50Mi"
+
+# Any custom router annotations(https://github.com/deis/router#annotations)
+# which needs to be applied can be specified as key-value pairs under "deployment_annotations"
+deployment_annotations:
+  #<example-key>: <example-value>
+
+# Any custom annotations for k8s services like http://kubernetes.io/docs/user-guide/services/#ssl-support-on-aws
+# which needs to be applied can be specified as key-value pairs under "service_annotations"
```
mboersma: Same comment as above--since "annotations" is plural, the verb should be `need` not `needs`. Nitpick++. 😄

rimusz:
LGTM, just wording mentioned by @mboersma needs to be fixed

PR #296, opened 2016-12-22
feat(charts): make hostPort and service type configurable

lachie83:


lachie83:
cc @krancour @slack @gabrtv

charts/router/values.yaml
```diff
@@ -15,3 +15,10 @@ deployment_annotations:
 # which need to be applied can be specified as key-value pairs under "service_annotations"
 service_annotations:
   #<example-key>: <example-value>
+
+# Enable to pin router pod hostPort when using minikube or vagrant
+host_port:
+  enabled: false
+
+# Service type default to LoadBalancer 
+service_type: NodePort
```
krancour: Given that the code in the template says:

```
type: {{ default "LoadBalancer" .Values.service_type}}
```

And given that the `values.yaml` line above isn't commented, the default value (if `values.yaml` is left untouched) is effectively `NodePort`, which I don't think is what you want. (Per the comment on line 23, I'm _certain_ that isn't what you intended.)

There are a few ways you could sort this out, but personally, I would change line 14 of the template to read:

```
type: {{ .Values.service_type }}
```

And change line 24 in `values.yaml` to express our intended default.

(The other option is to just add a comment in front of line 24 in the `values.yaml`, but for some reason I don't like that as much, but @lachie83 you might know better than me at this point what is considered the best practice or pattern for handling optional config and defaults.)

lachie83: Thanks for the review @krancour . That was sloppy on my part. I've commented it out in values and made LoadBalancer the default value in the comments

krancour:
Nice improvement.

PR #298, opened 2017-01-04
fix(charts): enable hostports by default

bacongobbler:
This was the old default behaviour and is expected. Changing it to disabled by default is a regression. change was introduced in #296

cc @lachie83 for visibility, does this impact you in any way?

### DIFF
--- a/charts/router/values.yaml
+++ b/charts/router/values.yaml
@@ -18,7 +18,7 @@ service_annotations:
 
 # Enable to pin router pod hostPort when using minikube or vagrant
 host_port:
-  enabled: false
+  enabled: true
 
 # Service type default to LoadBalancer 
 # service_type: LoadBalancer


bacongobbler:
this actually doesn't fix the problem, one sec

bacongobbler:
there, that should resolve it and charts that flag it with `enabled: false` to work as well.

vdice:
[ci](https://ci.deis.io/job/workflow-chart-e2e/372/console) is hitting a template render error:
```
16:38:28 Installing workflow chart workflow-v2.9.1-20170104233738-sha.cc7dcfa
16:38:29 Error: render error in "workflow/charts/router/templates/router-deployment.yaml": template: workflow/charts/router/templates/router-deployment.yaml:52:7: executing "workflow/charts/router/templates/router-deployment.yaml" at <eq .Values.host_port...>: error calling eq: incompatible types for comparison
```

lachie83:
Confirming that this doesn't impact me. I'm fine with the default being enabled.

lachie83:
@bacongobbler appreciate you getting a fix in.

bacongobbler:
Ah so the template originally worked, I'm just being dumb because I thought it was a string rather than a boolean value. Things should be working again. :)

@lachie83 not a problem, thanks for the PR! I know lots of people have asked for something like this.

PR #299, opened 2017-01-09
[WIP] - feat(*): add configurable server_tokens and default route

lachie83:
Resurrecting this patch https://github.com/deis/router/compare/deis:master...gerred:server-tokens

-  feat(*): add configurable server_tokens nginx setting
- Conditional default service enabled

### DIFF
--- a/nginx/config.go
+++ b/nginx/config.go
@@ -48,6 +48,9 @@ http {
 
 	client_max_body_size {{ $routerConfig.BodySize }};
 
+	{{ if $routerConfig.DisableServerTokens -}}
+	server_tokens off;
+	{{- end}}
 	{{ range $realIPCIDR := $routerConfig.ProxyRealIPCIDRs -}}
 	set_real_ip_from {{ $realIPCIDR }};
 	{{ end -}}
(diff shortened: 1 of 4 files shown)


bacongobbler:
@lachie83 when CI has finished testing (pass or fail) you can test with `quay.io/deisci/router:git-71e35df`.

nginx/config.go
```diff
@@ -128,31 +131,6 @@ http {
...
-		location ~ ^/healthz/?$ {
-			access_log off;
-			default_type 'text/plain';
-			return 200;
-		}
-		location / {
-			return 404;
-		}
-	}
-
```
jeroenvisser101: @lachie83 we might want to add this back when DefaultServiceEnabled is false, since removing the default might break healthchecks on existing clusters?

lachie83: Yes this is the code I'm a little worried about. Your suggestion makes sense.

lachie83:
Moved server_token configuration to #301

lachie83:
Moved default app to #302

lachie83:
Split this work out into two PRs. Closing

PR #303, opened 2017-01-17
fix(charts): remove deployment and service annotations

bacongobbler:
closes https://github.com/deis/workflow/issues/689

### DIFF
--- a/charts/router/values.yaml
+++ b/charts/router/values.yaml
@@ -8,12 +8,12 @@ dhparam: ""
 
 # Any custom router annotations(https://github.com/deis/router#annotations)
 # which need to be applied can be specified as key-value pairs under "deployment_annotations"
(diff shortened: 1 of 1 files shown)


charts/router/values.yaml
```diff
@@ -6,16 +6,6 @@ dhparam: ""
 # limits_cpu: "100m"
 # limits_memory: "50Mi"
 
-# Any custom router annotations(https://github.com/deis/router#annotations)
-# which need to be applied can be specified as key-value pairs under "deployment_annotations"
-deployment_annotations:
```
mboersma: Should we just comment out `deployment_annotations:` here rather than remove it?

bacongobbler: Yeah that might be better.

Issue #293, opened 2016-11-29
need annotation for nginx proxy_buffer_size


jshimko:
If an upstream app has a header that's bigger than the default proxy buffer size, nginx returns a 502 and the client receives nothing.

Example logs...

```sh
2016/11/29 18:10:10 [error] 64#0: *59606 upstream sent too big header while reading response header from upstream, client: 100.96.2.1, server: ~^subdomain\.(?<domain>.+)$, request: "GET / HTTP/1.1", upstream: "http://101.69.46.220:80/", host: "subdomain.example.com"

[2016-11-29T18:10:10+00:00] - 100.96.2.1 - - - 502 - "GET / HTTP/1.1" - 732 - "-" - "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_12_1) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/57.0.2936.0 Safari/537.36" - "~^subdomain\x5C.(?<domain>.+)$" - 101.69.46.220:80 - subdomain.example.com - 0.005 - 0.005
```

The explanation/fix...
https://ma.ttias.be/nginx-proxy-upstream-sent-big-header-reading-response-header-upstream/

So, essentially just need to be able to add those proxy buffer lines to the config...

```
server {
  listen        80;
  server_name   host.tld;

  location / {
    proxy_pass       http://upstream;
    ...

    proxy_buffer_size          128k;
    proxy_buffers              4 256k;
    proxy_busy_buffers_size    256k;
  }
}
```

Reference:

http://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_buffer_size

gemoya:
Do you have a expected relase date with this changes??

jshimko:
Thanks again @krancour!

krancour:
No problem, @jshimko.

vdice:
@gemoya these changes are now included in the latest Router release, [v2.11.0](https://github.com/deis/router/releases/tag/v2.11.0).  Testing will commence shortly for the next Workflow release to include this version (Workflow v2.12) and we expect it will be available by tomorrow.

PR #307, opened 2017-01-27
feat(nginx): Allow proxy buffer size/number configuration.

Anichale:
addresses: #293 

Adds proxy_buffers, proxy_buffer_size, proxy_busy_buffers_size lines to the nginx config in config.go along with matching Model amendments to AppConfig and default nginx values.

Edit: Only added configuration for the routable-application component, as I'm not sure if you guys want the deis-router and deis-builder proxy_buffers configurable. Also, have yet to add anything to the list of Annotations in the readme, but I will if you give the ok.


bacongobbler:
It looks like there's some tabs/spaces conflicting in some of the files causing some indentation errors. Would you mind fixing those up? Thanks!

bacongobbler:
Jenkins, test this please.

vdice:
Jenkins, test this please.

vdice:
@Anichale seeing the following on the latest CI run for this/these changes:
```
2017/01/30 22:00:03 INFO: Starting nginx...
2017/01/30 22:00:03 INFO: nginx started.
2017/01/30 22:00:03 INFO: Router configuration has changed in k8s.
2017/01/30 22:00:03 INFO: Reloading nginx...
2017/01/30 22:00:03 INFO: nginx reloaded.
nginx: [emerg] "proxy_buffers" directive invalid value in /opt/router/conf/nginx.conf:163
```
(as a result, the router pod never enters a running state)

Artifact used: `quay.io/deisci/router:git-3c67ce5` if needed to debug/replicate issue...

Anichale:
Ah, my mistake. I'll pick this up later.

vdice:
@Anichale Do you have access to the Travis build results?  (https://travis-ci.org/deis/router/builds/198899466) Seeing the following from `make test-cover`:
```
--- FAIL: TestValidAppProxyBufferSize (0.00s)
	model_validation_test.go:387: Using value "1|2", received an unexpected error: Field "proxyBufferSize" value "1|2" does not satisfy constraint /^[1-9]\d*[kKmM]?$/
FAIL
coverage: 29.2% of statements
FAIL	github.com/deis/router/model	0.039s
make: *** [test-cover] Error 1
```

vdice:
Jenkins, test this please.

krancour:
I understand the impetus for adding this feature, but there are some implementation specifics here that need to be addressed.

First, and entirely independent of the rest of my feedback, it's not clear to me from my reading of Nginx docs, but are `proxy_buffers` and `proxy_buffer_size` mutually exclusive options? Or should they be? It seems `proxy_buffers` allows you to configure _number and size_ of buffers, while `proxy_buffer_size` allows you to configure just size. Let's make sure we're crystal clear on whether it's meaningful to use both of these options together or whether we can get by with just the `proxy_buffers` and maybe two separate annotations like `router.deis.io/nginx.proxyBuffers.number` and `router.deis.io/nginx.proxyBuffers.size`.

__tl;dr on the rest: Rename annotations to:__

* `router.deis.io/nginx.proxyBuffer.number`
* `router.deis.io/nginx.proxyBuffer.size`
* `router.deis.io/nginx.proxyBuffer.busySize`

Here's why:

1. I would recommend that as with ssl, gzip, and a few other things that are general configuration _categories_, with each possessing a number of more specific options within, the annotations should look more like `router.deis.io/proxyBuffer.<option>`. Likewise, the relevant parts of the model should be compartmentalized in a `ProxyBufferConfig` type. That should all be fairly easy to address.

1. My second bit of feedback is more nuanced. The configuration options (currently) supported by the router have been very carefully segregated into options that are nginx-specific (or more accurately, specific to _our_ nginx-based router implementation) and options that are _not_ implementation-specific.
\
Currently, all the annotations that are applicable to the router's deployment object are implementation-specific (you will notice all such annotations are prefixed with `router.deis.io/nginx.`) and all the annotations on the routable services are _not_ nginx-specific. (To be fair, the documentation for some of those options mentions Nginx, but it shouldn't.) This latter category of options all refer to _general_ concepts that any theoretical router implementation should be able to work with: domains, certificates, timeouts, etc. i.e. These annotations are the basis of a de-facto spec for other router implementations. (Some do exist in the wild.)
\
The options you're adding, however, are nginx-specific-- and not because of the similarity between their names and corresponding Nginx directives, but because the semantics of _how_ a proxy buffer is configured _in Nginx_ (defined in terms of the three very specific options you are exposing) is bleeding through. Think about an alternative router implementation based on Apache or Vulcan, for instance. While general concepts like domain, certificate, or timeout all survive the leap from one implementation to another intact, do _specific_ options like `proxyBusyBuffersSize` mean anything in those other contexts?
\
I see three possible paths forward.

    1. The first is appealing, but seems prohibitively difficult. It would require somehow distilling proxy buffer configuration options into more generic or universal terms that could be interpreted and applied by any router implementation.

    1. The second is that these options move up to the router's deployment resource and join the ranks of all the existing nginx-specific configuration options. This is probably unappealing because those are all _global_ options and wouldn't give you fine-grained control over proxy buffer size for individual applications-- which, I think, is something you want.

    1. The middle ground is that we start allowing implementation-specific options on the routable services. There isn't any existing precedent for this, but I see no problem with establishing that as long as the annotations are renamed appropriately-- i.e. prefixed with `router.deis.io/nginx.`. (Don't forget all other renaming guidance given earlier in this comment.) I think it would make it clear that these are _not_ part of the de-facto spec that alternative router implementations should follow. The caveat to this, however, is that being implementation-specific, do not expect that the deis CLI or Workflow controller will ever evolve to manage this particular set of new options. These annotations (like a few others) would have to be applied to the routable service manually by an operator.

Does all of that make sense? I know this all sounds very academic, but I think it's important to carefully preserve the option to implement alternative routers.

Anichale:
Makes sense, and thanks for taking the time to write out explanations.

I'll go ahead and rewrite the annotations/nginx config in lines with the third path you specified.

As `proxy_buffers` and `proxy_buffer_size` are not mutually exclusive - size input from `proxy_buffers` overrides `proxy_buffer_size` -  I'll only include proxy_buffers as a configurable field.

krancour:
@Anichale did you happen to have the changes we discuss queued up? Internally, this is something we want to move along sooner rather than later. I stand ready to finish up what you've started, but just want to avoid stepping on toes. I'll be getting to work on it Tues. morning (EST) unless I hear that I shouldn't.

krancour:
#323 handled this and has been merged.

PR #311, opened 2017-02-07
feat(charts): add resource requests to chart

metadave:
- this allows the router to be used in a horizontal pod autoscaler out the the box. Without `resources.requests`, the HPA can't calculate a `CURRENT` value

### DIFF
--- a/charts/router/templates/router-deployment.yaml
+++ b/charts/router/templates/router-deployment.yaml
@@ -32,8 +32,9 @@ spec:
       - name: deis-router
         image: quay.io/{{.Values.org}}/router:{{.Values.docker_tag}}
         imagePullPolicy: {{.Values.pull_policy}}
-{{- if or (.Values.limits_cpu) (.Values.limits_memory)}}
+{{- if or (.Values.limits_cpu) (.Values.limits_memory) (.Values.requests_cpu) (.Values.requests_memory)}}
         resources:
+{{- if or (.Values.limits_cpu) (.Values.limits_memory)}}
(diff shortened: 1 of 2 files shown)


deis-admin:

mboersma:
Could you amend the commit message to follow the [commit style guide](https://deis.com/docs/workflow/contributing/submitting-a-pull-request/#commit-style)? That way it will get picked up by `deisrel` when we generate the release changelog.

Jenkins, add to whitelist.

metadave:
fixed, thank you @mboersma

charts/router/templates/router-deployment.yaml
```diff
@@ -32,16 +32,28 @@ spec:
       - name: deis-router
         image: quay.io/{{.Values.org}}/router:{{.Values.docker_tag}}
         imagePullPolicy: {{.Values.pull_policy}}
-{{- if or (.Values.limits_cpu) (.Values.limits_memory)}}
+{{- if or (.Values.limits_cpu) (.Values.limits_memory) (.Values.requests_cpu) (.Values.requests_memory)}}
```
krancour: `or` can take more than two arguments?

metadave: I believe so: https://github.com/golang/go/blob/master/src/text/template/funcs.go#L295

mboersma:
I've tested this manually and it looks like the `resources:` stanza is created correctly when need be.

metadave:
thank you @mboersma!

PR #316, opened 2017-02-14
feat(ingress) Experimental Native Ingress

kris-nova:
Preliminary work for disabling the router chart when using `EXPERIMENTAL_NATIVE_INGRESS`

Note: this will probably be refactored once we figure out #771

Requires deis/builder#495
Requires deis/workflow#732
Requires deis/controller#1243


krancour:
@kris-nova, I think I see a problem here...

`.Values.global.experimental_native_ingress` looks like a value from the "parent" Workflow chart, which merely depends upon this chart. I don't think this chart will ever see that value (i.e., it can't look "up"), which can certainly complicate the notion of conditionally installing or not installing the resources defined in this chart based on some config option in the parent chart.

Thinking through this more, I suspect what we _really_ need to do is not touch the router chart at all, but _conditionally_ include it or not include it in the Workflow chart. I'm wondering if that can actually be done... going to consult with the Helm team a bit.

.gitignore
```diff
@@ -4,3 +4,4 @@ rootfs/opt/router/sbin/router
 rootfs/opt/router/sbin/router.*
 vendor/
 coverage.txt
+.idea*
```
mboersma: Hiding files under .idea isn't strictly related to the ingress changes, so ideally this would be in a separate PR.

charts/router/templates/router-deployment.yaml
```diff
@@ -76,3 +77,4 @@ spec:
             port: 9090
           initialDelaySeconds: 1
           timeoutSeconds: 1
+{{ end }}
```
mboersma: These files should end with a newline to satisfy some tools.

Would it improve readability to finish with a comment like:
```yaml
{{ end }}{{/* if not .Values.global.experimental_native_ingress */}}
```
I'm just asking--not sure myself.

kris-nova: Added, also added a newline back to the end.. dang emacs 😝

.gitignore
```diff
@@ -1,4 +1,4 @@
 rootfs/opt/router/sbin/router
 rootfs/opt/router/sbin/router.*
 vendor/
-coverage.txt
+coverage.txt
```
krancour: Looks like this removes the trailing newline in this file. Can we leave that be?

charts/router/templates/router-deployment.yaml
```diff
@@ -1,3 +1,4 @@
+{{- if not .Values.global.experimental_native_ingress }}
```
krancour: I think I may have said something to this effect once before, but I don't think this works... `.Values.global` looks like a reference to the "top level" configuration in the Workflow chart. This chart can't "look up" into there because... well... imagine the scenario where this chart is installed standalone (i.e. without the rest of Workflow)... there wouldn't even be anything to look up _to_.

I _think_ this repo probably does not require any changes to support the native ingress initiative.

kris-nova: I added the `global.experimental_native_engress` value to the `values.yaml` for the router chart.. this should support a standalone installation of the router

krancour: Ah. I am sorry for a bit of mis-understanding here. I'd forgotten that `global` was a special scope within Helm that allowed charts to "look up." I've now realized that...

I understand what you're trying to do here. Regardless of whether someone installs the router chart directly _or_ installs the Workflow chart, which includes the router chart, a value for `global.experimental_native_ingress` should be available.

Although I understand what you're trying to do and agree it would work, I think there are a couple smells here... and I think there's a much cleaner alternative.

The smells:

The router chart now has the ability, when installed, depending on the value of `global.experimental_native_ingress` to actually _not_ install any of its k8s resources. i.e. It is possible to effect a "no-op" install of the router. _(Install this chart-- but dont!)_ This doesn't make sense _unless_ considered in the broader context of re-use by the Workflow chart. That means the router chart has been designed with that specific re-use by the Workflow chart in mind. And that's a smell. Because, what really, does "experimental native ingress" mean to the router itself? Nothing, in fact.

As I've said before, I think the router requires no changes to support this experiment. (Neither Go nor chart changes are in order.) All that needs to be done is for the Workflow chart to _conditionally_ include or not include the router chart. Until re-reading the Helm documentation this evening, I'd hoped, but hadn't known for sure that was possible... it is...

This section of the documentation shows two techniques that could be used to conditionally include or exclude the router chart in its entirety from a Workflow installation:

https://github.com/kubernetes/helm/blob/master/docs/charts.md#tags-and-condition-fields-in-requirementsyaml

mboersma: Point: @knova I think I misunderstood the context earlier. I see @krancour's point that this makes a change to the router's own chart that allows it to be disabled completely, which seems wrong in the context of the router chart standing alone. We do intend the individual component charts to be useful separate from Workflow itself if at all possible.

Counterpoint: changing to use the new tags and conditions in Helm requires Workflow to specify `helm` v2.2.1 or later, is my understanding. And other components such as minio are disabled in a similarly inelegant way already. It could be argued that using tags + conditions isn't in scope for this overall feature, and that we should instead make a coordinated effort to refactor all conditional components at once.

@krancour would you be amenable to allowing this smelly-but-consistent change into router if we follow up with a set of PRs that embrace Helm tags & conditions that remove the current `{{ if`` blocks from all the Workflow charts?

kris-nova:
@mboersma @bacongobbler does anyone else see anything that needs attention on this one?

bacongobbler:
Nope! Looks like all of the templated objects from this chart is covered.

mboersma:
With the caveat that we intend to refactor this eventually (see deis/workflow#771), I'm ok with this change.

charts/router/templates/router-deployment.yaml
krancour: > @krancour would you be amenable to allowing this smelly-but-consistent change into router if we follow up with...

Sorry, @mboersma.., late to the game on this, but yet. 👍

mboersma: Thanks @krancour, no worries. We do intend to revert this change at some point in the future in favor of the cleaner tags & conditionals approach.

kris-nova: Sorry my code smells

PR #317, opened 2017-02-28
chore(Makefile): Remove minimally useful clean targets

krancour:
These two targets provide little or no value. I'd rather see them go away in the name of reducing the clutter in the Makefile.


mboersma:
Jenkins, test this please.

PR #318, opened 2017-02-28
chore(Makefile): Eliminate DEV_ENV_CMD_INT

krancour:
Minor thing...

Something I've learned since I first crafted this Makefile (and others like it for other repos)...

You don't _have_ to skip the `-it` when you don't care about a container running interactively and with a tty... and in fact, if you just always include the `-it`, we even get colorized output in places where we didn't previously.


krancour:
Welp... nevermind. Jenkins doesn't like it, because it's incapably of attaching a tty to the container. Closing.

PR #319, opened 2017-02-28
Eliminate legacy manifests and update hack / deploy process

krancour:
Router hasn't gotten as much love as some other Workflow components over the last few months... Where most repos shed their "legacy" `manifests/` directories some time ago, router still has one. The hack / build / deploy process facilitated by `make deploy` depends on those very out-of-date manifests.

This PR puts a modernized process in place for "slotting" a hacked / custom-built image into a running cluster as a replacement for the stock router from the Workflow chart.

All docs are updated accordingly as well.

__Edit:__ Although this was all overdue, I also consider it a means to an end to enable me to finish #307


krancour:
@mboersma excellent question. I have no clue! Looking forward to @vdice's insight.

vdice:
@mboersma @krancour the new example charts in `charts/examples` should not affect the packaging of the "umbrella" Workflow chart.  

When the aforementioned chart is assembled/updated it reaches out to existing component charts as hosted in their respective chart repos.  To that end, I've also reminded myself that the job for assembling/publishing a component chart operates only on/from the `charts/<component name>` sub-directory, which applies here for router as well.

That being said, we can make sure of both statements once ci runs (that is, proceeds past current Dockerhub/Quay/S3-related issues).  It will package/publish a router pr chart and kick off a downstream workflow pr chart job.  We can `helm fetch --untar router-pr/router --version <version as used in ci>` to be sure the `examples` are not included.

mboersma:
> once ci runs

They tell me patience is a virtue, but that just makes me want to punch them. 😛  

[Looks to the sky and yells]
ESSSS-THREEEEE!!!

krancour:
@vdice CI ran... does all look as you expect, re: examples charts not bleeding into anywhere it ought not?

vdice:
@krancour indeed -- all looks good!  Sanity check:
```
$ helm repo add router-pr https://charts.deis.com/router-pr
...
$ helm fetch --untar router-pr/router --version v2.10.1-20170301165338-sha.3e8ff74
$ tree router/
router/
├── Chart.yaml
├── templates
│   ├── router-deployment.yaml
│   ├── router-dhparam-secret.yaml
│   ├── router-service-account.yaml
│   └── router-service.yaml
└── values.yaml

1 directory, 6 files
```

PR #321, opened 2017-02-28
fix(.travis.yml): Make travis apply style checks

krancour:
This makes travis apply style checks. It's a pretty big oversight that it isn't currently.

This also includes minor formatting fixes to `.travis.yml` itself.

Note: this will fail until #320 is merged.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,6 +13,6 @@ env:
 install:
 - make bootstrap
 script:
-  - make build test-cover
+- make build test-style test-cover
 after_success:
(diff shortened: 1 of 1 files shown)


vdice:
LGTM -- thoughts on adding `test-unit` as well?  (No `test*` targets are currently run by the Jenkins-based build/deploy jobs)

krancour:
@vdice the `test-unit` runs unit tests without code coverage. `test-cover`, which is already being invoked here runs unit tests _with_ coverage. So, as far as unit tests are concerned, we're... _covered_ here. 😆

Issue #326, opened 2017-03-02
Proposal: Add aws load balancer timeout override to service template


vdice:
As AWS still seems to be the most popular k8s cluster provider for installing Workflow, seemingly even for dev clusters, perhaps the annotation mentioned in https://deis.com/docs/workflow/managing-workflow/configuring-load-balancers/#idle-connection-timeouts should be added to the default router service [template](https://github.com/deis/router/blob/master/charts/router/templates/router-service.yaml) so that operators don't bump into issues stemming from the default load balancer timeout of 60s, as seems to be the frequent case currently.

Thoughts?

krancour:
Agree. I actually implemented that feature upstream in k8s with the intention of making this same modification to the chart... but then I forgot  about it.

Cryptophobia:
This issue was moved to teamhephy/router#11

PR #327, opened 2017-03-03
feat(chart): add aws-specific idle timeout annotation to router service

vdice:
Ref https://github.com/deis/router/issues/326


krancour:
For anyone who may have this concern, please note that this AWS-annotation will not have any adverse effect when this chart is installed elsewhere.

bacongobbler:
:shipit: ?

PR #328, opened 2017-03-06
fix(charts): quote annotation values

### LINKED ISSUE #325: Trouble setting service_annotations via chart values file
I've recently attempted setting various `service_annotations` for the router service via a `values.yaml` -- both via Workflow's global version and Router's.  The former errors out while the latter appears successful yet the annotations don't show up in resulting service resource:

Via Workflow's `values.yaml`:

Relevant snippet:
```
router:
  dhparam: ""
  # Any custom router annotations(https://github.com/deis/router#annotations)
  # which need to be applied can be specified as key-value pairs under "deployment_annotations"
  deployment_annotations:
    #<example-key>: <example-value>

  # Any custom annotations for k8s services like http://kubernetes.io/docs/user-guide/services/#ssl-support-on-aws
  # which need to be applied can be specified as key-value pairs under "service_annotations"
  service_annotations:
    router.deis.io/nginx.proxyBuffers.number: "10"
    router.deis.io/nginx.proxyBuffers.size: "5k"
    router.deis.io/nginx.proxyBuffers.busySize: "10k"
    router.deis.io/connectTimeout: "35s"
```
Helm install attempt:
```
$ helm install --namespace deis --name workflow workflow-dev/workflow -f values.yaml
2017/03/02 14:16:26 warning: destination for service_annotations is a table. Ignoring non-table value <nil>
Error: release workflow failed: Service in version "v1" cannot be handled as a Service: [pos 196]: json: expect char '"' but got char '1'
```

Via Router's `values.yaml`:

Relevant snippet:
```
# Any custom annotations for k8s services like http://kubernetes.io/docs/user-guide/services/#ssl-support-on-aws
# which need to be applied can be specified as key-value pairs under "service_annotations"
service_annotations:
  router.deis.io/nginx.proxyBuffers.number: "10"
  router.deis.io/nginx.proxyBuffers.size: "5k"
  router.deis.io/nginx.proxyBuffers.busySize: "10k"
  router.deis.io/connectTimeout: "35s"
```

Helm install is successful but when we check out the `deis-router` service resource we don't see the annotations intended to be set:
```
$ helm install --namespace deis --name workflow workflow/
...
$ kd get svc deis-router -o yaml
apiVersion: v1
kind: Service
metadata:
  creationTimestamp: 2017-03-02T21:18:17Z
  labels:
    heritage: deis
  name: deis-router
  namespace: deis
  resourceVersion: "19624"
  selfLink: /api/v1/namespaces/deis/services/deis-router
  uid: c0d2148e-ff8d-11e6-ba37-1644177bcf32
spec:
  clusterIP: 10.0.0.240
  ports:
  - name: http
    nodePort: 31264
    port: 80
    protocol: TCP
    targetPort: 8080
  - name: https
    nodePort: 30161
    port: 443
    protocol: TCP
    targetPort: 6443
  - name: builder
    nodePort: 30178
    port: 2222
    protocol: TCP
    targetPort: 2222
  - name: healthz
    nodePort: 31671
    port: 9090
    protocol: TCP
    targetPort: 9090
  selector:
    app: deis-router
  sessionAffinity: None
  type: LoadBalancer
status:
  loadBalancer: {}
```

_Note: some of the attempted annotations are related to PR https://github.com/deis/router/pull/323_

Helm version v2.2.1

mboersma:
I get errors trying to set `service_annotations` or `deployment_annotations` in testing #323. Templating the router service seems to be the initial problem, because the `range` loop doesn't quote the values as required.

Closes #325.

### DIFF
--- a/charts/router/templates/router-deployment.yaml
+++ b/charts/router/templates/router-deployment.yaml
@@ -7,7 +7,7 @@ metadata:
   annotations:
     component.deis.io/version: {{ .Values.docker_tag }}
 {{- range $key, $value := .Values.deployment_annotations }}
-    {{ $key }}: {{ $value }}
+    {{ $key }}: {{ quote $value }}
 {{- end }}
 {{- if not (empty .Values.platform_domain) }}
     router.deis.io/nginx.platformDomain: {{ .Values.platform_domain }}
--- a/charts/router/templates/router-service.yaml
+++ b/charts/router/templates/router-service.yaml
@@ -5,7 +5,7 @@ metadata:
 {{- if .Values.service_annotations }}
   annotations:
 {{- range $key, $value := .Values.service_annotations }}
-    {{ $key }}: {{ $value }}
+    {{ $key }}: {{ quote $value }}
 {{- end }}
 {{- end }}
   labels:


charts/router/templates/router-deployment.yaml
```diff
@@ -7,7 +7,7 @@ metadata:
   annotations:
     component.deis.io/version: {{ .Values.docker_tag }}
 {{- range $key, $value := .Values.deployment_annotations }}
-    {{ $key }}: {{ $value }}
+    {{ $key }}: "{{ $value }}"
```
krancour: I think there's a `quote` function in sprig for this that might be safer it will escape any double quote that exist within the string to make it YAML safe.

mboersma: Good point--I'll try `| quote`.

bacongobbler:
I'll see if I can get service annotations working with this fix.

bacongobbler:
This worked for me. LGTM!

```
><> cat workflow/values.yaml | grep proxyBuffers
    router.deis.io/nginx.proxyBuffers.enabled: "true"
><> kd get deploy deis-router -o yaml | grep proxy
    router.deis.io/nginx.proxyBuffers.enabled: "true"
```

PR #330, opened 2017-03-07
ref(router): disable router hostPort usage by default

slack:
Refs https://github.com/deis/workflow/pull/757


krancour:
Duplicate of #329-- closing as #329 has the advantage of already having passed CI.

PR #333, opened 2017-03-30
Add status endpoint

pixeleet:
I'd like to add a sysdig app chcek on the router, but it only installs the stub_status module, doesn't use it to expose the endpoint or provides configuration for it. 

@krancour What else would I need to submit to get this in? Is this even correct at all?

### DIFF
--- a/nginx/config.go
+++ b/nginx/config.go
@@ -145,6 +145,12 @@ http {
 			return 200;
 		}
 
+    {{ if $routerConfig.statusEndpoint }}
+    location /nginx_status {
+      stub_status on;
+    }
+    {{ end }}
+
 		location / {
 			proxy_buffering {{ if $routerConfig.ProxyBuffersConfig.Enabled }}on{{ else }}off{{ end }};
 			proxy_buffer_size {{ $routerConfig.ProxyBuffersConfig.Size }};
--- a/nginx/config.go
+++ b/nginx/config.go
@@ -186,6 +186,11 @@ http {
 		location / {
 			return 404;
 		}
+		{{ if $routerConfig.statusEndpoint }}
+		location /nginx_status {
+			stub_status on;
+		}
+		{{ end }}
 	}
 	{{ end }}
 
--- a/nginx/config.go
+++ b/nginx/config.go
@@ -144,13 +144,13 @@ http {
 			default_type 'text/plain';
 			return 200;
 		}
-
     {{ if $routerConfig.statusEndpoint }}
     location /nginx_status {
       stub_status on;
+      allow 127.0.0.1;
+      deny all;
     }
     {{ end }}
-
 		location / {
 			proxy_buffering {{ if $routerConfig.ProxyBuffersConfig.Enabled }}on{{ else }}off{{ end }};
 			proxy_buffer_size {{ $routerConfig.ProxyBuffersConfig.Size }};
@@ -194,7 +194,9 @@ http {
 		}
 		{{ if $routerConfig.statusEndpoint }}
 		location /nginx_status {
-			stub_status on;
+      stub_status on;
+      allow 127.0.0.1;
+      deny all;
 		}
 		{{ end }}
 	}
--- a/model/model.go
+++ b/model/model.go
@@ -65,6 +65,7 @@ type RouterConfig struct {
 	HTTP2Enabled             bool                `key:"http2Enabled" constraint:"(?i)^(true|false)$"`
 	LogFormat                string              `key:"logFormat"`
 	ProxyBuffersConfig       *ProxyBuffersConfig `key:"proxyBuffers"`
+	StatusEndpoint           bool                `key:"statusEndpoint" constraint:"(?i)^(true|false)$"`
 }
 
 func newRouterConfig() (*RouterConfig, error) {
@@ -95,6 +96,7 @@ func newRouterConfig() (*RouterConfig, error) {
 		HTTP2Enabled:             true,
 		LogFormat:                `[$time_iso8601] - $app_name - $remote_addr - $remote_user - $status - "$request" - $bytes_sent - "$http_referer" - "$http_user_agent" - "$server_name" - $upstream_addr - $http_host - $upstream_response_time - $request_time`,
 		ProxyBuffersConfig:       proxyBuffersConfig,
+		StatusEndpoint:           false,
 	}, nil
 }
 


deis-admin:

nginx/config.go
```diff
@@ -293,6 +293,17 @@ http {
 		proxy_pass {{$builderConfig.ServiceIP}}:2222;
 	}
 }{{ end }}
+
+{{ if $routerConfig.StatusEndpoint }}
+  sevrer {
```
krancour: Type here.

pixeleet: good catch

krancour: Also, probably this shouldn't be its own vhost / server block. As it stands, I don' think any traffic will ever reach this endpoint. The router uses hostname-based routing... so every request either goes to an application _or_ to the default vhost. This new server block is neither of those things. Probably it would be better to add this to the existing default vhost. Once you do, the only way to access this endpoint will be by making a request to the router that doesn't match any other application-- e.g. `bogus.example.com/nginx_status`. Is that suitable?

Or I'm wondering if maybe we shouldn't reserve the `router-status` subdomain (relative to your platform domain) for use by the router (i.e. add it to the short list of names that the controller doesn't allow for user apps). Then router status could consistently be accesses at `router-status.example.com`...

We'd probably want to enable whitelists on that endpoint as well.

mboersma:
@krancour opined:
> Probably it would be better to add this to the existing default vhost...

This PR still has some outstanding review questions and isn't ready quite yet, so I'm going to move it to the next milestone.

nginx/config.go
pixeleet: Hm, I think it would be easy to say that the router can assume a domain for itself, that could be the default vhost, if the `statusEndpoint` option is `true` there we could expose it under the default `/nginx_status`. Is that a viable approach? 

I imagine it would make some things easier, like explaining how to hit the router from outside. But now this is becoming a new thing => Adding a router domain, or something along those lines, && enabling status endpoint on default vhost if I'm not mistaken.

krancour: Unless @mboersma objects, I am unopposed to `router.<domain>` (or maybe `deis-router.<domain>`) becoming the router's official address, much as `deis.<domain` is the address of the controller and `deis-builder.<domain>` is the address of the builder. We'd have to add `router` (or `deis-router`) to the list of reserved names in the controller.

That being said... that can be implemented independently (i.e. in another PR) of this feature. `router.<domain>` will _already_ get you to the efault vhost today _unless_ you have a userland app called `router`. Giving the router its own domain is just a formality of sorts. So, while I am unopposed, let's consider it on its own merits. Would you mind opening an issue?

In the meantime, this PR can proceed just by moving the `/nginx_status` endpoint to the existing default vhost.

pixeleet:
Added a new PR per the discussion with @krancour moved the endpoint into the default vhost and removed the restrictions on it as it should be managed by the whitelist held by the router

nginx/config.go
```diff
@@ -186,6 +186,9 @@ http {
 		location / {
 			return 404;
 		}
+		location /nginx_status {
+			stub_status on;
+		}
```
krancour: You still want this to be conditional though, yes? Also, should this `location` (but not the entire default vhost) be either locked down to a configurable whitelist, or possibly protected with a configurable username/password. I tend to imagine we don't want to expose these stats to the world.

pixeleet: I'll read up on whitelisting in the router, pushed a single commit for the configurability

krancour:
@pixeleet, your last commit seems to have a lot of (breaking) changes in it that appear to be beyond the scope of this PR. Accidental?

nginx/config.go
```diff
@@ -61,7 +58,7 @@ http {
 	real_ip_header X-Forwarded-For;
 	{{- end }}
 
-	log_format upstreaminfo '{{ $routerConfig.LogFormat }}';
+	log_format upstreaminfo '[$time_iso8601] - $app_name - $remote_addr - $remote_user - $status - "$request" - $bytes_sent - "$http_referer" - "$http_user_agent" - "$server_name" - $upstream_addr - $http_host - $upstream_response_time - $request_time';
```
pixeleet: I'm not sure how this got here

pixeleet:
I cleaned it up into a single commit, and will look into what we talked about whitelisting or a basic auth addition to it.

nginx/config.go
```diff
@@ -186,6 +186,11 @@ http {
 		location / {
 			return 404;
 		}
+		{{ if $routerConfig.statusEndpoint }}
```
krancour: I see you groomed the commits and force-pushed... I think you lost the model changes that complement this.

pixeleet: Damn indeed. :) That's stupid.

krancour:
@pixeleet until #334 is addressed, for good measure, the same `location` block should be conditionally added [right here](https://github.com/deis/router/blob/master/nginx/config.go#L147) to account for the case that someone has provided an application for use as the default vhost.

nginx/config.go
```diff
@@ -145,6 +145,12 @@ http {
 			return 200;
 		}
 
+    {{ if $routerConfig.statusEndpoint }}
+    location /nginx_status {
+      stub_status on;
+    }
+    {{ end }}
+
```
krancour: Can the leading whitespace be made consistent with what's above and below?

pixeleet: Sure, just a quick question, are you fine with just restricting /nginx_status to 127.0.0.1 for now and deny all? the use case primarily being monitoring this should be fine

nginx/config.go
krancour: I would consider moving this section up between the health check and the `location /` block. I know Nginx works by making the best/closest match, but for human readability, it usually makes the most sense to have the more specific patterns listed first so that you can think of unmatched requests as progressively "falling through" the list for a possible match to a more general pattern... if you don't mind.

nginx/config.go
krancour: Yes. We've already done that very same thing here: https://github.com/deis/router/blob/master/nginx/config.go#L202-L207

This makes me wonder a few things.

- Are you aware that lots of router metrics are already available in Grafana? (Just making sure.)
- Are the metrics that are already exposed at `127.0.0.1:9090/stats` something you can work with? i.e. Is this new endpoint needed for something you can't already get?
- Purely to satisfy my own curiosity, what agent / metric collector would you be running that would be hitting this from the router's own loopback interface?

pixeleet: 1, yes I'm aware, we're using mainly sysdig and the grafana part we skipped from our setup
2, sysdig wants me to have the nginx stub status exposed
3, sysdig cloud for the containers and newrelic for api performance metrics
4, 127.0.0.1:9090/nginx_status works just fine for me! :)

krancour: > 127.0.0.1:9090/nginx_status works just fine for me! 

Ok. Awesome. Let's start there. If, as time goes on, there is some compelling reason to expose this beyond that, we can take it up then.

krancour:
@pixeleet I think, probably based on our conclusion of putting this endpoint at `127.0.0.1:9090/nginx_status`, there is no remaining need to make this conditional. Thoughts?

pixeleet:
Sounds good to me :)

pixeleet:
I'll close this PR and will open a separate clean one just for the `/nginx_status` location on 9090

PR #338, opened 2017-04-25
fix(charts): remove b64enc from dhparam

mpon:
I found that the dhparam value was applied two times base64 encoding.
I think it is unnecessary here, written in README below following.

> Include the base64 encoded dhparam in a secret:


krancour:
This appears to address a legitimate bug that looks to have been introduced in the transition from helm classic charts to helm 2.0+ charts. @mboersma or @bacongobbler can either of you just double-check this?

mboersma:
Yep, looks like an old bug to me as well. This makes it agree with the docs, as well as the other router params that are required to be pre-encoded with `base64`. Thanks @mpon!

Issue #274, opened 2016-10-06
Optimize the control loop


krancour:
Currently, the router does _all_ of the following  every ten seconds:
1. Find the router's own deployment object.
2. Find all routable services.
3. Find secrets that contain certificates

All of these are the inputs used to build / compute a model. This model is deep compared to the previously computed model, which is kept in memory. When there are differences, Nginx configuration is re-generated by using the new model as input to a template. Nginx is reloaded with the new configuration.

Note that the model is computed _before_ any comparison is made so that inconsequential changes to k8s resources (those that wouldn't affect the router's configuration) won't trigger an unnecessary Nginx reload.

Here's where there's room to optimize-- I feel like continuously computing the model is wasteful since changes that affect the router's configuration are a relatively rare event. This not only wastes CPU cycles, but depending on how many routable services live in your cluster, this process can be _very_ chatty with the apiserver. This puts an unnecessary load on the network.

A more mature approach to this may be to watch the k8s event stream instead. Only in the event that a k8s resource we're interested in (as determined by a label or well known name) has been added or modified should we care about retrieving those resources and re-computing the model. What's more, because we'd know _what_ changed before re-computing the model, we could also re-compute and _replace_ a portion of the existing model.

cc @arschles -- I feel like if there's any obvious problem with this approach, you'd be the guy to spot it. 😉

bacongobbler:
related issue that this may resolve: #212

krancour:
@bacongobbler I'm not sure about it solving #212... the reason is that when the router first starts, it needs to compute the _entire_ config model. It's _after_ that that we can watch the event stream and just compute deltas.

bacongobbler:
Ah yeah I guess the base case wouldn't be resolved, but rather every subsequent case. Good call.

krancour:
@arschles and I both felt this would work. He can speak to it better than I, but he did say something about possible problems with the k8s client disconnecting. There's reconnect logic in Steward that covers that base. Would have to do the same here.

krancour:
Experience working on the Kubernetes Service Catalog's controller has taught me that what we need to implement here is the so-called "informer pattern," which is the common pattern implemented by Kubernetes controllers. (The router, essentially, _is_ a Kubernetes controller.)

Cryptophobia:
This issue was moved to teamhephy/router#17

PR #344, opened 2017-05-07
RBAC support

Bregor:
With this change `deis-router` became available to work in **RBAC**-only clusters

Works with both Kubernetes 1.5 and 1.6 (see `templates/_helpers.tmpl` for details)
Actually tested with 1.5.7 and 1.6.2

Role allows `deis-router`:
- `extensions/deployments`: `get`
- `apps/deployments`: `get`

ClusterRole allows `deis-router`:
- `services`: `get` and `list`
- `secrets`: `get`
- `endpoints`: `get`

### DIFF
--- a/charts/router/templates/_helpers.tmpl
+++ b/charts/router/templates/_helpers.tmpl
@@ -1,0 +1,10 @@
+{{/*
+Set apiVersion based on Kubernetes version
+*/}}
+{{- define "rbacAPIVersion" -}}
+{{- if ge .Capabilities.KubeVersion.Minor "6" -}}
+rbac.authorization.k8s.io/v1beta1
+{{- else -}}
+rbac.authorization.k8s.io/v1alpha1
+{{- end -}}
+{{- end -}}
--- a/charts/router/templates/router-clusterrole.yaml
+++ b/charts/router/templates/router-clusterrole.yaml
@@ -1,0 +1,21 @@
+{{- if (.Values.global.use_rbac) -}}
+{{- if (.Capabilities.APIVersions.Has (include "rbacAPIVersion" .)) -}}
+kind: ClusterRole
+apiVersion: {{ template "rbacAPIVersion" . }}
+metadata:
+  name: deis:deis-router
+  labels:
+    app: deis-router
+    heritage: deis
+rules:
+- apiGroups: [""]
+  resources: ["services"]
+  verbs: ["get", "list"]
+- apiGroups: [""]
+  resources: ["secrets"]
+  verbs: ["get"]
+- apiGroups: [""]
+  resources: ["endpoints"]
+  verbs: ["get"]
+{{- end -}}
+{{- end -}}
--- a/charts/router/templates/router-clusterrolebinding.yaml
+++ b/charts/router/templates/router-clusterrolebinding.yaml
@@ -1,0 +1,19 @@
+{{- if (.Values.global.use_rbac) -}}
+{{- if (.Capabilities.APIVersions.Has (include "rbacAPIVersion" .)) -}}
+kind: ClusterRoleBinding
+apiVersion: {{ template "rbacAPIVersion" . }}
+metadata:
+  name: deis:deis-router
+  labels:
+    app: deis-router
+    heritage: deis
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: deis:deis-router
+subjects:
+- kind: ServiceAccount
+  name: deis-router
+  namespace: {{ .Release.Namespace }}
+{{- end -}}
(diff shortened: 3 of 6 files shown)


vdice:
Jenkins, OK to test

vdice:
Jenkins, OK to test

charts/router/templates/_helpers.tmpl
```diff
@@ -0,0 +1,10 @@
+{{/*
+Set apiVersion based on Kubernetes version
+*/}}
+{{- define "rbacApiVersion" -}}
```
krancour: If observe Go's established naming conventions, this should be `rbacAPIVersion`. See doc [here](https://github.com/golang/go/wiki/CodeReviewComments#interfaces).

charts/router/templates/router-clusterrole.yaml
```diff
@@ -0,0 +1,22 @@
...
+apiVersion: {{ template "rbacApiVersion" . }}
+metadata:
+  name: deis:deis-router
+  labels:
+    app: deis-router
+    heritage: deis
+rules:
+- apiGroups: ["extensions", "apps"]
+  resources: ["deployments"]
+  verbs: ["get", "list", "watch"]
```
krancour: The only deployment that the router is concerned with is its own. It's retrieved by name-- i.e. a get. Could you please try removing list and watch here and verify that the router still works?

charts/router/templates/router-clusterrole.yaml
```diff
@@ -0,0 +1,22 @@
...
+- apiGroups: [""]
+  resources: ["services"]
+  verbs: ["get", "list", "watch"]
```
krancour: There's an issue open (https://github.com/deis/router/issues/274) that would require watch for services, but in the current implementation, watch isn't used (but list is). Could you please try removing it here and verify that the router still works? On principle, we should be giving the minimum set of permissions required.

charts/router/templates/router-clusterrole.yaml
```diff
@@ -0,0 +1,22 @@
...
+- apiGroups: [""]
+  resources: ["secrets"]
+  verbs: ["get", "list"]
```
krancour: Secrets are retrieved by name-- i.e. a get-- so list shouldn't be needed. Could you please try removing list here and verify that the router still works?

charts/router/templates/router-clusterrole.yaml
```diff
@@ -0,0 +1,22 @@
+{{ if .Capabilities.APIVersions.Has (include "rbacApiVersion" .) }}
+kind: ClusterRole
+apiVersion: {{ template "rbacApiVersion" . }}
+metadata:
+  name: deis:deis-router
```
krancour: Is this colon allowed here? Even if so, I would consider changing this name. My understanding is that names of most things in k8s should look like valid DNS names, which I believe excludes colons. Even if it's technically permissible in this particular spot, I would rather we still deferred to that convention.

krancour:
@Bregor thank you for this contribution. I made several comments, but I want to tell you that overall, this looks great!

charts/router/templates/router-clusterrole.yaml
Bregor: @krancour of course it is. More than that: it's recommended for "global" (not namespaced) resources.
Look at clusterroles, defined by kubernetes itself:

    $ kubectl get clusterrole
    NAME                                           AGE
    ...
    system:auth-delegator                          1h
    system:basic-user                              6h
    system:controller:attachdetach-controller      1h
    system:controller:certificate-controller       1h
    system:controller:cronjob-controller           1h
    system:controller:daemon-set-controller        1h
    system:controller:deployment-controller        1h
    system:controller:disruption-controller        1h
    system:controller:endpoint-controller          1h
    system:controller:generic-garbage-collector    1h
    system:controller:horizontal-pod-autoscaler    1h
    system:controller:job-controller               1h

Bregor:
@krancour thank you very much for your comments! I will try to achieve them in couple hours.

charts/router/templates/_helpers.tmpl
Bregor: @krancour thank you, naming updated.

vdice:
Jenkins, add to whitelist

Bregor:
@krancour thank you once again for you comments and thank you for too long waiting.
Fixed and rebased.

krancour:
@Bregor did your own testing show that router still worked fine with the reduced permissions?

Bregor:
@krancour I still can create/delete app, set domains an so on.

krancour:
Thanks, @Bregor. LGTM. Waiting for one other LGTM.

PR #346, opened 2017-05-24
docs(README): update charts link/reference

### LINKED ISSUE #339: Documentation still has some references to Helm Classic

vdice:
Closes #339


README.md
```diff
@@ -487,13 +487,13 @@ __This configuration is not suitable for production.__ The primary use case for
 
 ### Customizing the charts
 
-The Helm Classic charts available for installing router (either with or without the rest of Deis Workflow) are intended to get users up and running as quickly as possible.  As such, the charts do not strictly require any editing prior to installation in order to successfully bootstrap a cluster.  However, there are some useful customizations that should be applied for use in production environments:
+The Helm [chart][] available for installing router (either with or without the rest of Deis Workflow) are intended to get users up and running as quickly as possible.  As such, the charts do not strictly require any editing prior to installation in order to successfully bootstrap a cluster.  However, there are some useful customizations that should be applied for use in production environments:
```
mboersma: There's just one router chart for kubernetes/helm, right? So I think the first sentence should use "is" instead of "are," and the second sentence should change to:
> As such, the chart does not strictly require...

vdice: ah thank you, good call; updated.

PR #347, opened 2017-06-12
Allow returning 200 for root requests

tmc:
This enables placing deis-router behind a GCE ingress which sends health-checks to ```/```.


deis-admin:

mboersma:
This LGTM--thanks @tmc for doing this. @krancour could you take a look at this too?

README.md
```diff
@@ -249,6 +249,7 @@ _Note that Kubernetes annotation maps are all of Go type `map[string]string`.  A
 | <a name="default-service-enabled"></a>deis-router | deployment | [router.deis.io/nginx.defaultServiceEnabled](#default-service-enabled) | `"false"` | Enables default back-end service for traffic hitting /. In order to work correctly both `defaultServiceIP` and `DefaultAppName` MUST also be set.  |
 | <a name="default-app-name"></a>deis-router | deployment | [router.deis.io/nginx.DefaultAppName](#default-app-name) | `""` | Default back-end application name for traffic hitting router on /. In order to work correctly both `defaultServiceIP` and `DefaultServiceEnabled` MUST also be set.  |
 | <a name="default-service-ip"></a>deis-router | deployment | [router.deis.io/nginx.defaultServiceIP](#default-service-ip) | `""` | Default back-end service ip for traffic hitting router on /. In order to work correctly both `DefaultAppName` and `DefaultServiceEnabled` MUST also be set. |
+| <a name="return-200-at-root"></a>deis-router | deployment | [router.deis.io/nginx.return200AtRoot](#return-200-at-root) | `"false"` | Whether or not to turn a 200 for otherwise unhandled requests to `/` |
```
krancour: "Whether or not to turn a 200" should be "Whether or not to _return_ a 200"

krancour:
I'm not opposed to this, but I'm not entirely clear on why you'd want to do this. The router is, essentially, an ingress controller (although it uses annotated services instead of actual ingress resources because it was created at a time before ingress was stable). Workflow also has support now for "bring your own ingress controller," and you wouldn't include a router in your Workflow installation if you used that option. What I'm not understanding is why one would opt to put an ingress controller _in front of_ the router, since I see those two components as mutually exclusive.

mboersma:
Jenkins, add to whitelist.

mboersma:
> Workflow also has support now for "bring your own ingress controller,"

True, but until some issues are fixed that support is incomplete.

> why one would opt to put an ingress controller in front of the router...?

@tmc could you elaborate a bit on the use case for this?

mboersma:
@tmc we still have some questions about the general usefulness of this change. I'm going to close it since it hasn't gotten attention in a while, but please re-open with some comments if I was too hasty.

Thanks again!

